PR #15, opened 2015-02-20
Hadoop on lustre2

ehiggs:
Configs that support Hadoop-on-lustre2 and Hadoop-2.5.0-cdh5.3.1 using GPFS.

These versions of Hadoop require the native libraries to be built and an EasyBuild block/config is in the works.


etc/hod/Hadoop-2.5.0-cdh5.3.1-gpfs/hod.conf
```diff
@@ -0,0 +1,62 @@
...
+#workdir=
+directories=$localworkdir/dfs/name,$localworkdir/dfs/data
+
+[core-site.xml]
+hadoop.tmp.dir=$localworkdir/tmp
+dfs.replication=1
+fs.defaultFS=gpfs:///
+fs.gpfs.impl=org.apache.hadoop.fs.gpfs.GeneralParallelFileSystem
+fs.AbstractFileSystem.gpfs.impl=org.apache.hadoop.fs.gpfs.GeneralParallelFs
+gpfs.mount.dir=/user/scratchphanpy
```
stdweird: `phanpy`?

etc/hod/Hadoop-2.5.0-cdh5.3.1-gpfs/screen.conf
```diff
@@ -0,0 +1,13 @@
+#-*- cfg -*-
+# vim: ft=cfg
+
+[Unit]
+Name=screen
+RunsOn=master
+
+[Service]
+# We run 'exec bash' at the end so screen doesn't exit upon command completion.
+ExecStart=screen -dmS HODclient bash -c "module load Hadoop/2.4.0-seagate; HADOOP_CONF_DIR=$localworkdir/conf; exec bash"
```
stdweird: `2.4.0-seagte`? there's no varaible for this?

etc/hod/Hadoop-on-lustre2/screen.conf
```diff
@@ -0,0 +1,14 @@
+#-*- cfg -*-
+# vim: ft=cfg
+
+[Unit]
+Name=screen
+RunsOn=master
+
+[Service]
+module=Hadoop/2.3.0-cdh5.0.0
```
stdweird: is this correct?

stdweird:
minor reamrks, looks ok i guess (still don't like the large amounts of identical data between all the configs)

etc/hod/Hadoop-2.5.0-cdh5.3.1-gpfs/screen.conf
```diff
@@ -0,0 +1,14 @@
+#-*- cfg -*-
+# vim: ft=cfg
+
+[Unit]
+Name=screen
+RunsOn=master
+
+[Service]
+modules=Hadoop/2.5.0-cdh5.3.1
```
stdweird: `modules` or `module`? (same above too, but no templating there)

etc/hod/Hadoop-on-lustre2/screen.conf
```diff
@@ -0,0 +1,14 @@
+#-*- cfg -*-
+# vim: ft=cfg
+
+[Unit]
+Name=screen
+RunsOn=master
+
+[Service]
+modules=Hadoop/2.4.0-seagate
```
stdweird: same here: modules vs module

hod/config/config.py
```diff
@@ -89,8 +89,13 @@ def _parse_runs_on(s):
     options = dict(master=RUNS_ON_MASTER, slave=RUNS_ON_SLAVE, all=RUNS_ON_ALL)
     return options[s.lower()]
 
-def _cfgget(config, section, item, dflt=None):
-    '''Get a value from a ConfigParser object or a default if it's not there.'''
+def _cfgget(config, section, item, dflt=None, **kwargs):
```
stdweird: does this work? (named arg + the `**kwargs` magic)

hod/config/config.py
```diff
@@ -89,8 +89,13 @@ def _parse_runs_on(s):
...
+    '''
+    Get a value from a ConfigParser object or a default if it's not there.
+    Options in kwargs come from the command line and override the config.
```
stdweird: `from the command line` makes no sense here. just put `Options in kwargs override the config`

hod/config/config.py
ehiggs: It sure does!
```python
    In [1]: def foo(a, b, **kwargs):
       ...:     return a + b + sum([v for k,v in kwargs.items()])
       ...: 

    In [2]: foo(1, 2, x=3, **dict(y=4, z=5))
    Out[2]: 15
```

hod/config/config.py
```diff
@@ -117,11 +122,11 @@ class PreServiceConfigOpts(object):
 
     OPTIONAL_FIELDS=['master_env', 'modules', 'service_configs', 'directories']
 
-    def __init__(self, fileobj):
+    def __init__(self, fileobj, **kwargs):
         _config = load_service_config(fileobj)
         self.version = _cfgget(_config, _META_SECTION, 'version', '')
```
stdweird: why no `**kwargs` here too?

hod/hodproc.py
```diff
@@ -130,7 +131,8 @@ def distribution(self, *master_template_args, **kwargs):
         m_config_filenames = parse_comma_delim_list(m_config_filenames)
 
         self.log.info('Loading "%s" manifest config', m_config_filenames)
-        m_config = preserviceconfigopts_from_file_list(m_config_filenames)
+        m_config = preserviceconfigopts_from_file_list(m_config_filenames,
+                workdir=self.options.options.config_workdir)
```
stdweird: move the whole block that is common with `ConfiguredMaster` in new private method.

hod/config/config.py
stdweird: ok, just looks very hackish :smile:

hod/hodproc.py
ehiggs: Yeah I think there's a lot of repetition here but there's not really a semantic grouping I can see here. It's basically manually compressing code for the sake of the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

I'll take a look, though.

PR #16, opened 2015-03-02
[WIP] Autoconf

ehiggs:
Add hadoop config auto generation.

This is a work in progress. The added configs come from #15 .

The way to get the configs to be automatically generated is to add in the Config section of the `hod.conf`: `autogen=hadoop`. This lets us add future optional functionality by, for example, `autogen=hadoop,hdfs`

The autogenerated configs only update defaults. Users can then also override these values with their own `hod.conf` values by setting the values as previously.


hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+import os
+import re
+from hod.node.node import Node
+
+__all__ = ['autogen_config']
+
+def parse_memory(memstr):
+    '''
+    Given a string representation of memory, return the memory size in
+    bytes. It supports up to terabytes.
```
stdweird: can you give some examples here, because `kB` probably doesn't do what this function suggests

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+    '''
+    size = re.search(r'[bBkKmMgGtT]', memstr)
+    if size is not None:
+        coef = float(memstr[:size.start()])
```
stdweird: ?

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+        return coef * (1024 ** 2)
+    elif exp in ['g', 'G']:
+        return coef * (1024 ** 3)
+    elif exp in ['t', 'T']:
+        return coef * (1024 ** 4)
+    raise RuntimeError('Unable to parse memory amount:', memstr)
+
+def format_memory(mem, round_val=False):
+    '''
+    Given an integer 'mem', return the string with associated units Note
```
stdweird: `.` before `Note`

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+    that this is used to set heap sizes for the jvm awhich doesn't accept non
```
stdweird: `awhich` either typo (or me not known english very well ;)

hod/config/autogen/hadoop.py
stdweird: what units is `mem` suppsoed to be in? bytes i assume?

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+                return '%d%s' % (mem_in_units, units[unit_idx])
+        unit_idx -= 1
+    return '%db' % mem
+
+# mapping for total system memory -> reserved for OS
+# Retrieved from:
+# http://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.0.6.0/bk_installing_manually_book/content/rpm-chap1-11.html
+MemRec = namedtuple('MemRec', ['total','os'])
+_RECOMMENDATIONS = [
+    MemRec(parse_memory('8g'), parse_memory('2g')),
```
stdweird: can you generate this list, arther then copy/paste it?

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+    MemRec(parse_memory('128g'), parse_memory('24g')),
+    MemRec(parse_memory('256g'), parse_memory('32g')),
+    MemRec(parse_memory('512g'), parse_memory('64g')),
+]
+
+def reserved_memory(totalmem):
+    for mem in _RECOMMENDATIONS:
+        if totalmem <= mem.total:
+            return mem.os
+    return _RECOMMENDATIONS[-1].os
```
stdweird: default is largest? please add comment why.

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+    MemRec(parse_memory('64g'), parse_memory('8g')),
+    MemRec(parse_memory('72g'), parse_memory('8g')),
+    MemRec(parse_memory('96g'), parse_memory('12g')),
+    MemRec(parse_memory('128g'), parse_memory('24g')),
+    MemRec(parse_memory('256g'), parse_memory('32g')),
+    MemRec(parse_memory('512g'), parse_memory('64g')),
+]
+
+def reserved_memory(totalmem):
+    for mem in _RECOMMENDATIONS:
```
stdweird: docstring

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+        if totalmem <= mem.total:
+            return mem.os
+    return _RECOMMENDATIONS[-1].os
+
+def min_container_size(totalmem):
+    if totalmem < parse_memory('4G'):
```
stdweird: docstring

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+        return  256 * (1024**2)
+    elif totalmem < parse_memory('8G'):
+        return 512 * (1024**2)
+    elif totalmem < parse_memory('24G'):
+        return 1024**3
+    else:
+        return 2 * (1024**3)
+
+MemDefaults = namedtuple('MemDefaults',
```
stdweird: class constants before functions

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+        [
+            'available_memory',
+            'min_container_sz',
+            'num_containers',
+            'ram_per_container'
+        ])
+
+def memory_defaults(total_memory, ncores):
+    available_memory = total_memory - reserved_memory(total_memory)
```
stdweird: docstrings also all otehr functions that have none

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,257 @@
...
+def blocksize(path):
+    try:
+        return os.statvfs(path).f_bsize
+    except OSError, e:
+        if e.errno == errno.ENOENT:
+            return os.statvfs(dirname(path)).f_bsize
+        raise
+
+
+def set_default(d, key, val):
```
stdweird: why this function?

hod/config/autogen/hadoop.py
ehiggs: kb works fine. I'll put in a test and doctest to make it clear.

hod/config/autogen/hadoop.py
ehiggs: If it finds the unit then the coeficient is `memstr[0:beginning_of_unit_string]`

hod/config/autogen/hadoop.py
ehiggs: I don't think there is a function to generate the list that will be clearer than the expanded list.

hod/config/autogen/hadoop.py
ehiggs: If sets a value if it doesn't already exist. If I'm having a brain fart and forgetting a builtin that does this, let me know.

ehiggs:
Recent patches add support for detecting use of a partial node and scaling the memory appropriately.

hod/config/autogen/common.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def available_memory(node):
+    '''
+    Return the amount of memory available in bytes. There are three things we
+    consider: 
+    1. If we are using all the cores, we assume we can use all the
+    memory in the machine (minus what the OS needs).
+
+    2. If ulimit is set and less than meminfo, we use it, we use it.
```
stdweird: `we use it` once

hod/config/autogen/common.py
```diff
@@ -0,0 +1,217 @@
...
+    '''
+    for mem in _RECOMMENDATIONS:
+        if totalmem <= mem.total:
+            return mem.os
+    # totalmem > 512g
+    return _RECOMMENDATIONS[-1].os
+
+
+def available_memory(node):
+    '''
```
stdweird: can you add a todo here or an issue to support cgroups and parsing pbs environment variables?

hod/config/autogen/common.py
```diff
@@ -0,0 +1,217 @@
...
+    memory = meminfo - reserved_memory(meminfo)
+    # If we have the whole box, let's use all the non OS memory
+    if node['cores'] == node['totalcores']:
+        return memory
+
+    pct_cores = float(node['cores']) / node['totalcores']
+
+    memory = int(memory * pct_cores)
+    ulimit = node['memory']['ulimit']
+    if ulimit == 'unlimited':
```
stdweird: actually, unlimited is an alias for some value in some kernel header file. it's certainly not unlimited

hod/config/autogen/common.py
```diff
@@ -0,0 +1,217 @@
...
+def set_default(d, key, val):
+    '''If a value is not in dict d, set it'''
+    if key not in d:
+        d[key] = val
+    return d
+
+def update_defaults(d1, d2):
+    '''Update dict d1 with data from d2 iff values are not already in d1.'''
+    for k, v in d2.items():
+        set_default(d1, k, v)
```
stdweird: use `d1.setdefault(k,v)`?

hod/config/autogen/common.py
```diff
@@ -0,0 +1,217 @@
...
+
+    memory = int(memory * pct_cores)
+    ulimit = node['memory']['ulimit']
+    if ulimit == 'unlimited':
+        return int(memory)
+    else:
+        return min(int(memory), int(ulimit))
+
+def set_default(d, key, val):
+    '''If a value is not in dict d, set it'''
```
stdweird: it's confusing with regualr `setdefault` (i also don't see why you'd return the dict `d`. normal setdefault return `d[key]`).

hod/config/autogen/common.py
```diff
@@ -0,0 +1,217 @@
...
+    elif mem_in_mb > 1024:
+        denom = 256
+    else:
+        denom = 128
+
+    return int(math.floor(mem_in_mb/denom) * denom)
+
+def cap(val, limit):
+    '''
+    Return val unless val is over the limit; then return the limit.
```
stdweird: like `min(val,limit)`?

hod/config/autogen/common.py
```diff
@@ -0,0 +1,217 @@
...
+    to the nearest unit where mem will be over 1.
+
+    Note that this is used to set heap sizes for the jvm which doesn't accept
+    non integer sizes. Therefore this truncates.
+
+    Note also that this supports outputting 'b' for bytes even though java
+    -Xmx${somenum}b won't work.
+    '''
+    units = 'bkmgt'
+    unit_idx = len(units) - 1
```
stdweird: hmm, not sure if it this more readable:
```python
>>> units = 'bkmgt'
>>> for idx,unit in list([(idx,v) for idx, v in enumerate(units)])[::-1]: print idx,unit
... 
4 t
3 g
2 m
1 k
0 b

```

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,166 @@
...
+MemDefaults = namedtuple('MemDefaults', [
+    'available_memory',
+    'min_container_sz',
+    'num_containers',
+    'ram_per_container'
+])
+
+def min_container_size(totalmem):
+    '''
+    Given an amount of memory in bytes, return theh amount of memory that
```
stdweird: `theh` -> `the`

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,166 @@
...
+    represents the minimum amount of memory (in bytes) to be allocated in a Yarn
```
stdweird: if it's truly specific to yarn, add it iin the function name.

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,166 @@
...
+    }
+    return dflts
+
+def mapred_site_xml_defaults(workdir, node_info):
+    '''
+    Default entries for the mapred-site.xml config file.
+    '''
+    mem_dflts = memory_defaults(node_info)
+
+    java_map_mem = format_memory(0.8 * mem_dflts.ram_per_container, round_val=True)
```
stdweird: any reference where the `0.8` comes from?

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,166 @@
...
+    java_reduce_mem = format_memory(0.8 * 2 * mem_dflts.ram_per_container, round_val=True)
```
stdweird: smae here (magic factors)

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,166 @@
...
+    map_memory = cap(round_mb(mem_dflts.ram_per_container), 8192)
+    reduce_memory = cap(round_mb(2 * mem_dflts.ram_per_container), 8192)
+    dflts = {
+        'mapreduce.framework.name': 'yarn',
+        'mapreduce.map.java.opts': '-Xmx%s' % java_map_mem,
+        'mapreduce.map.memory.mb': map_memory,
+        'mapreduce.reduce.java.opts': '-Xmx%s' % java_reduce_mem,
+        'mapreduce.reduce.memory.mb': reduce_memory,
+        # io.sort.mb can't be > 2047mb
+        'mapreduce.task.io.sort.mb': min(int(0.4 * map_memory), 2047),
```
stdweird: more magic factors

hod/config/autogen/hadoop.py
```diff
@@ -0,0 +1,166 @@
...
+    '''
+    Default entries for the mapred-site.xml config file.
+    '''
+    mem_dflts = memory_defaults(node_info)
+
+    java_map_mem = format_memory(0.8 * mem_dflts.ram_per_container, round_val=True)
+    java_reduce_mem = format_memory(0.8 * 2 * mem_dflts.ram_per_container, round_val=True)
+    # In my tests, Yarn gets shirty if I try to run a job and these values are set to
+    # more then 8g:
+    map_memory = cap(round_mb(mem_dflts.ram_per_container), 8192)
```
stdweird: move the `8192` to some module constant (to be used in next function)

hod/config/autogen/hadoop_on_lustre2.py
```diff
@@ -0,0 +1,78 @@
...
+from hod.config.autogen.common import update_defaults
+
+__all__ = ['autogen_config']
+
+def core_site_xml_defaults(workdir, node_info):
+    dflts = {
+        'hadoop.tmp.dir': '$workdir/tmp'
+    }
+    base_hadoop_dflts = hcah.core_site_xml_defaults(workdir, node_info)
+    return update_defaults(dflts, base_hadoop_dflts)
```
stdweird: is this not the same as
```python
base_hadoop_dflts = hcah.core_site_xml_defaults(workdir, node_info)
base_hadoop_dflts.update(dflts)
return base_hadoop_dflts
```

hod/config/autogen/hadoop_on_lustre2.py
```diff
@@ -0,0 +1,78 @@
...
+
+def yarn_site_xml_defaults(workdir, node_info):
+    dflts = {
+        'yarn.nodemanager.local-dirs': '$workdir/$hostname',
+    }
+    base_hadoop_dflts = hcah.yarn_site_xml_defaults(workdir, node_info)
+    return update_defaults(dflts, base_hadoop_dflts)
+
+def autogen_config(workdir, node_info):
```
stdweird: it's a pity this is not class based, becasue this is copy/paste from above

hod/node/node.py
```diff
@@ -167,13 +172,47 @@ def get_memory():
...
+    memory['meminfo'] = _get_memory_proc_meminfo()
+    memory['ulimit'] = _get_memory_ulimit_v()
     return memory
 
+def get_totalcores():
+    '''Parse /proc/cpuinfo to find the total number of cores.'''
+    cores = 0
+    proc_cpuinfo_filename = '/proc/cpuinfo'
+    cpu_info = open(proc_cpuinfo_filename, 'r').read()
+    cpu_info = cpu_info.replace(' ', '').replace('\t', '').split('\n')
```
stdweird:
```python
import re
return len(re.findall(r'^\s*processor\s*:', cpu_info ,re.M)) or 1
```
(i'd return 1 out of safety instead of null, to save you a few div by zeroes

stdweird: or use multiprocessing
```python
import multiprocessing
return multiprocessing.cpu_count()
```

hod/config/autogen/common.py
ehiggs: The string we see is 'unlimited' and in this case we just ignore the ulimit as a source of memory capping.

hod/config/autogen/common.py
ehiggs: tbh I could clone `d2` and call `d2.update(d1)`

hod/config/autogen/common.py
ehiggs: +1 removed.

hod/config/autogen/common.py
ehiggs: +1 removed.

hod/config/autogen/hadoop.py
ehiggs: It's in the `autogen/hadoop.py` module so it should be clearly the scope of hadoop/yarn containers rather than, e.g. docker or rocket. Do you think it's necessary to add yarn to the function name?

hod/config/autogen/hadoop.py
ehiggs: There is a reference in common.py to [this list of recommendations](http://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.0.6.0/bk_installing_manually_book/content/rpm-chap1-11.html).

hod/config/autogen/hadoop_on_lustre2.py
ehiggs: If this was class based, here is what it would look like:

```python
def yarn_site_xml_defaults(self, workdir, node_info):
  dflts = {
  'yarn.nodemanager.local-dirs': '$workdir/$hostname',
  }
  base_hadoop_dflts = super(HadoopOnLustre2, self).yarn_site_xml_defaults(workdir, node_info)
  return update_defaults(dflts, base_hadoop_dflts)
```
I don't see a palpable advantage.

hod/node/node.py
ehiggs: `multiprocessing.cpu_count()` calls `os.sysconf('SC_NPROCESSOR_ONLN') which calls libc's `sysconf` which calls [`__get_cpu_count`](https://github.com/lattera/glibc/blob/a2f34833b1042d5d8eeb263b4cf4caaea138c4ad/sysdeps/unix/sysv/linux/getsysstats.c#L127) which parses the input of `/sys/devices/system/cpu/online`. There is another function, [`__get_cpu_count_conf`](https://github.com/lattera/glibc/blob/a2f34833b1042d5d8eeb263b4cf4caaea138c4ad/sysdeps/unix/sysv/linux/getsysstats.c#L237) that reads the `/sys/devices/system/cpu/` directory and counts the number of entries with `cpu` in it. So it's kinda similar to this function anyway.

I guess I could just call `os.sysconf('SC_NPROCESSORS_ONLN')

stdweird:
@ehiggs can you rebase?

ehiggs:
Rebased.

PR #17, opened 2015-03-27
Strip configs (as they can now be autogenerated) and tidy up logging imports.

ehiggs:


stdweird:
so only 2 config files are affected by the autoconfig stuff? that seems odd?

ehiggs:
We smushed all the config into one file per distribution and we support Vanilla and 'ln -s' flavours of Hadoop 2.

stdweird:
but configs like https://github.com/hpcugent/hanythingondemand/blob/master/etc/hod/Hadoop-2.5.0-cdh5.3.1-gpfs/hod.conf can't be reduced as well?

ehiggs:
I reduced it a bit but it doesn't have its own generator so the config isn't completely gone.

PR #23, opened 2015-04-22
Module fixes.

ehiggs:
1. Add ability to load modules from the command line using
`--config-modules=comma,separated,list`
This allows us to use something like Hadoop-2.3.0 configs but then also load up Spark and Python by just passing `--config-modules=Python/2.7.9-intel-2015a,Spark/1.3.0` on the `hod_pbs.py` command line. This is required if we want the `nodemanager` to be able to find an appropriate Python to run when using `pyspark` without having to hack configs.

2. Add `$modules` as a template param for services to use.

3. Fix screen session so it reloads the loaded modules to fix the fact that
screen strips the `LD_` environment variables. This allows us to load things like Python in the screen session and it should actually work(!)

4. Bump version to 2.2.4

### DIFF
--- a/hod/config/config.py
+++ b/hod/config/config.py
@@ -115,18 +115,6 @@ def parse_comma_delim_list(s):
     '''
     return [x.strip() for x in filter(lambda x: x.strip(), s.split(','))]
 
-def _extend_list(lst, other):
-    '''Extend the list so 
-    _extend_list([1,2,3], 4) -> [1,2,3,4]
-    _extend_list([1,2,3], [4, 5]) -> [1,2,3,4, 5]
-    '''
-    if isinstance(other, list):
-        lst += other
-    else:
-        lst.append(other)
-    return lst
-
-
 class PreServiceConfigOpts(object):
     r"""
     Manifest file for the group of services responsible for defining service
@@ -155,7 +143,7 @@ class PreServiceConfigOpts(object):
             '''
             lst = parse_comma_delim_list(_cfgget(_config, _CONFIG_SECTION, name, ''))
             if kwargs.get(name, None) is not None:
-                _extend_list(lst, parse_comma_delim_list(kwargs[name]))
+                lst.extend(parse_comma_delim_list(kwargs[name]))
             return lst
 
         self.modules = _get_list('modules')
--- a/hod/config/config.py
+++ b/hod/config/config.py
@@ -98,7 +98,7 @@ def _cfgget(config, section, item, dflt=None, **kwargs):
     Get a value from a ConfigParser object or a default if it's not there.
     Options in kwargs override the config.
     '''
-    if item in kwargs and kwargs[item] is not None:
+    if kwargs.get(item, None) is not None:
         return kwargs[item]
     if dflt is None:
         return config.get(section, item)
@@ -154,7 +154,7 @@ class PreServiceConfigOpts(object):
             Mergely append.
             '''
             lst = parse_comma_delim_list(_cfgget(_config, _CONFIG_SECTION, name, ''))
-            if name in kwargs and kwargs[name] is not None:
+            if kwargs.get(name, None) is not None:
                 _extend_list(lst, parse_comma_delim_list(kwargs[name]))
             return lst
 
--- a/hod/config/config.py
+++ b/hod/config/config.py
@@ -115,6 +115,18 @@ def parse_comma_delim_list(s):
     '''
     return [x.strip() for x in filter(lambda x: x.strip(), s.split(','))]
 
+def _extend_list(lst, other):
+    '''Extend the list so 
+    _extend_list([1,2,3], 4) -> [1,2,3,4]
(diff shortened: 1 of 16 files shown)


hod/config/config.py
```diff
@@ -115,6 +115,18 @@ def parse_comma_delim_list(s):
     '''
     return [x.strip() for x in filter(lambda x: x.strip(), s.split(','))]
 
+def _extend_list(lst, other):
+    '''Extend the list so 
+    _extend_list([1,2,3], 4) -> [1,2,3,4]
+    _extend_list([1,2,3], [4, 5]) -> [1,2,3,4, 5]
+    '''
+    if isinstance(other, list):
+        lst += other
```
boegel: use `.extend`?

```python
if not isinstance(other, list):
    other = [other]
lst.extend(other)
return lst
```

hod/config/config.py
```diff
@@ -115,6 +115,18 @@ def parse_comma_delim_list(s):
...
+    else:
+        lst.append(other)
```
boegel: this will append to the originally passed list => side-effect!

hod/config/config.py
```diff
@@ -137,8 +149,14 @@ def _fixup_path(cfg):
...
 
         def _get_list(name):
-            return parse_comma_delim_list(_cfgget(_config, _CONFIG_SECTION,
-                name, '', **kwargs))
+            '''
+            With lists, we don't want to overwrite the value with kwargs.
+            Mergely append.
+            '''
+            lst = parse_comma_delim_list(_cfgget(_config, _CONFIG_SECTION, name, ''))
+            if name in kwargs and kwargs[name] is not None:
```
boegel:
```python
if kwargs.get(name, None) is not None:
```

hod/config/hodoption.py
```diff
@@ -69,6 +69,7 @@ def config_options(self):
 a comma separated list of config files with the later files taking
 precendence.""", "string", "store", ''),
                 'workdir': ("""Working directory""", "string", "store", None),
+                'modules': ("""Extra modules to load in each service environment""", "string", "store", None),
```
boegel: screwed up indent?

boegel:
just some minor remarks

I'm kind of surprised that you need to force reload those modules in a screen session though; maybe add a reference to some documentation explaining why `$LD_` is being fiddled with by `screen`?

boegel:
example reference: http://superuser.com/a/235773/141303

hod/config/hodoption.py
ehiggs: No. It lines up with the above lines.

hod/config/config.py
ehiggs: `lst.extend` would be side effectful, which you complain about in a later comment.

hod/config/config.py
ehiggs: Less clear, but more concise. I'll use your idea.

hod/config/config.py
stdweird: i'm with @boegel on this one. and i see no reason to have this in the first place

hod/config/config.py
```diff
@@ -137,8 +149,14 @@ def _fixup_path(cfg):
...
         def _get_list(name):
-            return parse_comma_delim_list(_cfgget(_config, _CONFIG_SECTION,
-                name, '', **kwargs))
+            '''
+            With lists, we don't want to overwrite the value with kwargs.
+            Mergely append.
+            '''
+            lst = parse_comma_delim_list(_cfgget(_config, _CONFIG_SECTION, name, ''))
+            if kwargs.get(name, None) is not None:
+                _extend_list(lst, parse_comma_delim_list(kwargs[name]))
```
stdweird: why not `lst.extend`?

hod/config/template.py
```diff
@@ -88,6 +97,7 @@ def register_templates(template_registry, workdir):
         ConfigTemplate('localworkdir', lambda: mklocalworkdir(workdir), 'Subdirectory of workdir with user, host, and pid in the name to make it distinct from other workdirs for use on shared file systems'),
         ConfigTemplate('user', _current_user, 'Current user'),
         ConfigTemplate('pid', os.getpid, 'PID for the current process'),
+        ConfigTemplate('modules', lambda: ' '.join(modules), 'Modules listed in the hod.conf'),
```
stdweird: what is this supposed to do?

ehiggs: It turns `$modules` into a space delimited string of the modules defined in the `hod.conf`. This is how something like `modules load` will want them. i.e. `module load Waffle Beer Chocolate`

stdweird: my issue was with the lamdab, but ii guess it expects something callable.

stdweird:
the diffs illustrate the need for global config files to handle config options that are the same cross version  (like the whole `screen.conf`).

anyway, maybe open an issue for that separately

PR #26, opened 2015-05-19
Use the hostaddress for the web app information. 

ehiggs:
The Hadoop web app uses the fqdn of the nodemanager or resource manager in the web app to make urls. This means that clicking around the web app you get links that will point to `node1234.mycluster` which probably only exists in your DNS for your cluster and not for the outside world (or even VPN). The result is that the links become effectively dead.

This patch tells hadoop about the IP address instead.Sadly Hadoop has ignores the fact that you provided an IP and inserts a FQDN instead. I'm talking on the user list to see if I should escalate this to a bug.

### DIFF
--- a/hod/mpiservice.py
+++ b/hod/mpiservice.py
@@ -133,8 +133,8 @@ def setup_tasks(svc):
         master_dataaddress = data_interface.addr
         fqdn = socket.getfqdn()
         master_template_kwargs = [
-                ConfigTemplate('masterhostname',fqdn,''),
-                ConfigTemplate('masterhostaddress',socket.gethostbyname(fqdn),''),
+                ConfigTemplate('masterhostname', fqdn,''),
+                ConfigTemplate('masterhostaddress', socket.gethostbyname(fqdn),''),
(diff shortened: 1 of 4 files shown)


hod/mpiservice.py
```diff
@@ -128,10 +128,15 @@ def setup_tasks(svc):
...
-        master_dataname = node.sorted_network(node.get_networks())[0].hostname
+        data_interface = node.sorted_network(node.get_networks())[0]
+        master_dataname = data_interface.hostname
+        master_dataaddress = data_interface.addr
+        fqdn = socket.getfqdn()
         master_template_kwargs = [
-                ConfigTemplate('masterhostname',socket.getfqdn(),''),
-                ConfigTemplate('masterdataname', master_dataname, '')
+                ConfigTemplate('masterhostname',fqdn,''),
+                ConfigTemplate('masterhostaddress',socket.gethostbyname(fqdn),''),
```
boegel: minor style remark: space before `,`

boegel:
looks fine, as far as I can tell

PR #27, opened 2015-05-22
Kickstart docs

boegel:
First version of HoD docs @ http://hod.readthedocs.org


ehiggs:
Remove .swp files.

docs/conf.py
```diff
@@ -0,0 +1,191 @@
...
+
+# General substitutions.
+project = 'hanythingondemand'
+copyright = '2015-2015, Ghent University, CC-BY-SA'
+
+# The default replacements for |version| and |release|, also used in various
+# other places throughout the built documents.
+#
+# The short X.Y version.
+version = '2.2.4'  # this is meant to reference the version of hanythingondemand
```
ehiggs: We should put all this stuff into a `setup.cfg` file and move it out of here and `setup.py`. Then load the cfg for these details so it's all in sync.

ehiggs:
Next release will be 2.2.5 so when this is merged and deployed it will be `2.2.5` so we may as well update it in the docs now.

docs/conf.py
boegel: You should discriminate between the version of the code, and the version covered by the documentation.

At some point, you may use `3.0dev` for the software version, indicating it's unstable, yet keep the docs at `2.x` (i.e. the latest stable release)

boegel:
No, don't mention a version in the docs that is not available yet.

When you issue the PR for 2.2.5, bump it in conf.py too.

ehiggs:
Version changes should be part of the CI deploy. hod doesn't have this yet and I don't imagine it will be tackled in this PR.

docs/conf.py
ehiggs: Good point. +1

PR #29, opened 2015-06-11
Add genconfig command for hod to create configuration files w/o a cluster

ehiggs:
Add genconfig command for hod to create configuration files w/o a cluster.

This allows us to inspect the resulting config on a node (or login node, I
guess) as though it was the master.

Issue #19, opened 2015-04-07
Tidy up command line processing


ehiggs:
Currently if you omit a required argument like `--config-config` then hod spits out some incoherent message that has nothing to do with the fact that the configuration file is missing. This needs to be fixed.

ehiggs:
#33

ehiggs:
This was fixed in #33.

boegel:
I disagree, I don't think we should consider this fixed until we've matched what we're outlining in #43.

boegel:
we've met the target set in #43, closing

PR #33, opened 2015-08-07
Command line handling fixes

ehiggs:
This introduces a few things. Maybe it should be broken up a bit...

1. Feature: Hadoop 2.5.0-cdh5.3.1-native configs for using native libraries (snappy, bzip2, zlib encoding in fetch for example).

2. Bug fix: Removes the use of `hwloc-bind socket:0` which I thought was removed ages ago.

3. Feature: Most substantially, it changes the command syntax to use `hod <subcommand>`. 

4. Bug Fix: `--config` as an option prefix has been removed. The options like `--config-config` are now just `--config`, `--modules`. `--workdir`.

5. Feature: Add `--dist` option that must be used without `--config`. `--dist` takes a name like `Hadoop-2.5.0-cdh5.3.1-native` and it looks up the file in the `etc/hod` directory and finds the appropriate `hod.conf` file. This will simplify the command line a great deal.


ehiggs:
hwloc-bind was indeed removed in #21 but somehow got back in through some bad rebase or something.

ehiggs:
This is missing the required documentation changes, but it's a bit PR so the review should begin asap.

ehiggs:
This depends on PR #29

bin/hod
```diff
@@ -0,0 +1,40 @@
...
+    print 'hod: hanythingondemand - Run services within a torque cluster ' 
+    print 'usage: hod [--version] [--help] <command> <command-args>'
+    print 'The most commonly used commands are:'
+    print '    create           Creates a cluster.'
+    print '    list             List running clusters.'
+    print '    help-template    Print template names and values for diagnostic purposes.'
+    print '    genconfig        Generate configuration files for a config.'
+    return 1
+
+cmds = {'create': pbs.CreatePbsApplication,
```
itkovian: Any reason you are not using GeneralOption? :)

hod/applications/genconfig.py
```diff
@@ -0,0 +1,61 @@
...
+
+@author: Ewan Higgs (Ghent University)
+"""
+import sys
+
+from textwrap import dedent
+
+from hod.applications.application import Application
+from vsc.utils import fancylogger
+_log = fancylogger.getLogger(fname=False)
```
itkovian: AFAIK, libraries should just import logging and use the logger that is created in the executable?

test/unit/applications/test_applications_genconfig.py
```diff
@@ -0,0 +1,43 @@
...
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+@author: Ewan Higgs (Universiteit Gent)
+"""
+
+import unittest
+import hod.applications.genconfig as hag
+
+from mock import Mock
```
itkovian: Unused import

bin/hod
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
+
+import sys
+
+try:
+    from mpi4py import MPI
+except ImportError, e:
+    print 'Error: Could not import mpi4py. Is the MPI environment module loaded?'
```
boegel: Which module for `MPI`?

bin/hod
```diff
@@ -0,0 +1,40 @@
...
+    print 'Undelying error is: "%s"' % e
```
boegel: typo: `Underlying`

bin/hod-local
```diff
@@ -0,0 +1,19 @@
...
+    from mpi4py import MPI
+except ImportError, e:
+    print 'Error: Could not import mpi4py. Is the MPI environment module loaded?'
+    print 'Undelying error is: "%s"' % e
+    sys.exit(1)
+
+from hod.applications import local
+
+def main(args):
+    app = local.LocalApplication()
```
boegel: I would strongly prefer only having a single command: `hod`, and support `hod --local`?

hod/applications/application.py
```diff
@@ -0,0 +1,36 @@
...
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+Base class for application commands.
+
+@author: Ewan Higgs (Universiteit Gent)
+"""
+
+
+class Application(object):
```
boegel: isn't `Application` a bit too generic (EasyBuild used to have a class like that too, but it was renamed to `EasyBlock` later).

Is `HODSubcommand` right (haven't gone through the whole thing yet).

hod/applications/genconfig.py
```diff
@@ -0,0 +1,61 @@
...
+#
+# hanythingondemand is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+"""
+hod-genconfig - parse a hod configuration file and output the resulting config
```
boegel: `hod genconfig` + `an hanything-on-demand configuration file`

hod/applications/local.py
```diff
@@ -39,22 +41,27 @@
...
+        s = """\
+        Run the hod program locally. Generally not run from the command line.
+        """
+        return dedent(s)
+
+    def run(self, args):
+        options = HodOption(go_args=args)
+
+        if MPI.COMM_WORLD.rank == MASTERRANK:
+            svc = ConfiguredMaster(options)
```
boegel: some debug logging here would be nice?

hod/rmscheduler/job.py
```diff
@@ -61,6 +61,8 @@ def submit(self):
     def generate_script(self):
         """Build the submit script"""
         script = ["#!/bin/bash", "## Generated by HOD"]
+        print 'script', script
+        print 'self', self
```
boegel: debug logging?

hod/rmscheduler/resourcemanagerscheduler.py
```diff
@@ -67,3 +67,4 @@ def header(self, nodes=5, ppn=-1, walltime=72):
         walltime = time in hours (can be float)
         """
         self.log.error("header not implemented")
+        return ""
```
boegel: log an error but still continue?

setup.py
```diff
@@ -62,7 +62,7 @@ def find_files(*dirs):
 
 PACKAGE = {
     'name': 'hanythingondemand',
-    'version': '2.2.4',
+    'version': '2.3.0',
```
boegel: I'd say this is 3.0.0, since the CLI has changed so much?

test/unit/applications/test_applications_genconfig.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+@author: Ewan Higgs (Universiteit Gent)
+"""
+
+import unittest
+import hod.applications.genconfig as hag
```
boegel: just import `GenConfigApplication` rather than introducing a cryptic rename like `hag`?

test/unit/applications/test_applications_helptemplate.py
```diff
@@ -0,0 +1,43 @@
...
+# You should have received a copy of the GNU General Public License
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+@author: Ewan Higgs (Universiteit Gent)
+"""
+
+import unittest
+import pytest
+import hod.applications.helptemplate as hah
```
boegel: haha

test/unit/applications/test_applications_pbs.py
```diff
@@ -0,0 +1,64 @@
...
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+@author: Ewan Higgs (Universiteit Gent)
+"""
+
+import unittest
+import pytest
+
+import hod.applications.pbs as hap
```
boegel: just import `CreatePbsApplication`

boegel:
@ehiggs: I don't have a good view on what's going on here, mostly because I'm not familiar with the codebase as is yet...

I'll try and get a better view on things tomorrow morning (while trying to tackle the docs update aspect that's missing here), but we should definitely sit together and see how things look like now as well.

hod/applications/application.py
```diff
@@ -0,0 +1,36 @@
...
+    '''Base class for application commands.'''
+    pass
+
```
boegel: Every subcommand should implement the `run` method, right? So, make this an abstract base class, with `run` an `abstractmethod`?

See https://docs.python.org/2/library/abc.html

bin/hod
ehiggs: Any. If you don't have mpi4py installed you get: 

```
Error: Could not import mpi4py. Is the MPI environment module loaded?
Undelying error is: "No module named mpi4py"
```

If you have mpi4py installed but not the underlying environment module, you get: 

```
Error: Could not import mpi4py. Is the MPI environment module loaded?
Undelying error is: "libmpi.so.1: cannot open shared object file: No such file or directory"
```

I think it's a helpful tip.

hod/applications/application.py
ehiggs: Subcommand might be more appropriate. I wouldn't use HODSubcommand since we have namespaces so we don't need a HOD prefix on everything.

hod/rmscheduler/resourcemanagerscheduler.py
ehiggs: All of these interface functions don't return anything. However in this case we want the default version to return a string since it's used to prepend to the generated script. Prepending `None` (which is the return value of a function that doesn't explicitly return anything) means we would otherwise need to say `if retval is None: retval = ''` in calling code. It's a bit dumb, imo.

setup.py
ehiggs: Sure. :)

test/unit/applications/test_applications_genconfig.py
ehiggs: I do this on all the tests since I could be importing a large number of classes and functions. I don't mind which is used but I think the consistency is important. I'll change it since I don't want to run afoul of Github's code of conduct. ;)

bin/hod
boegel: OK, but since there's no module named `MPI` typically, it's also partially confusing...

How about this, make it a single error message:

```python
sys.stderr.write("ERROR: Failed to import mpi4py: %s (is there a module loaded providing MPI support?)")
```

hod/applications/application.py
boegel: ok, let's go with Subcommand

hod/rmscheduler/resourcemanagerscheduler.py
boegel: But, then it's not an error to not have header implemented? You're just hitting the default implementation?

```python
def header(...):
    """..."""
    self.log.info("Returning empty header (default implementation of header())")
    return ''
```

test/unit/applications/test_applications_genconfig.py
boegel: +1 for consistency, but I see little point in prefixing all calls with `hag` in this test module (I didn't even realise the connotation here, it just seemed useless to me)

Making the `import` long because you're testing lots of classes/functions is fine, at least it's clear what's being tested then in the blink of an eye. :)

bin/hod
ehiggs: FWIW there is a module called mpi on fedora but I get your point.

bin/hod-local
ehiggs: I'm happy to use it as a subcommand, but we use GeneralOption's generate command which has no comprehension of the subcommands, so I don't think that would work.

I think the `--local` as an option makes no sense. Like should it work with other subcommands? (no). 

FWIW, some other programs using subcommand style just use the main script to exec the other subcommands. e.g.:

```
$ ls /usr/libexec/git-core/ | head
git
git-add
git-add--interactive
git-am
git-annotate
git-apply
git-archive
git-bisect
git-bisect--helper
git-blame
```
```
$ ls /usr/bin/hg*
/usr/bin/hg  /usr/bin/hg-ssh
```

So I don't share your strong preference.

boegel: I guess I'm not fully up to speed how `hod-local` is different from `hod`...

But now we're using a mixed approach: some are (only) subcommands (`genconfig`, `create`, ...), others are available as a seperate `hod-*` command (`hod-local`), which is confusing?

boegel:
Jenkins: test this please

bin/hod
```diff
@@ -0,0 +1,40 @@
...
+except ImportError, e:
+    sys.stderr.write('Error: Could not import mpi4py. Is there an environment module providing MPI support loaded?\n')
+    sys.stderr.write('Underlying error is: "%s"\n' % e)
+    sys.exit(1)
+
+from hod.subcommands import pbs, query, helptemplate, genconfig
+
+def usage():
+    print 'hod: hanythingondemand - Run services within a torque cluster ' 
+    print 'usage: hod [--version] [--help] <command> <command-args>'
```
boegel: `--version` doesn't work...

boegel:
Jenkins: test this please

ehiggs:
Jenkins: please test

boegel:
Jenkins: test this please

ehiggs:
Jenkins: test this please.

boegel:
This looks like an excellent step towards #19.

Some more changes should be made to the `hod` CLI, but I don't think blocking this PR for it is going to help us much in making good progress.

Let's merge this, and then continue in smaller isolated chunks.

I'll issue a separate PR to update the documentation with a good overview of the `hod` CLI (the one we want in the end).

boegel:
@ehiggs: please include https://github.com/ehiggs/hanythingondemand/pull/3, and then this is good to go

ehiggs:
Jenkins: test this please

ehiggs:
Good to go :shipit:

PR #43, opened 2015-08-13
add documentation on reworked HOD CLI (REVIEW)

boegel:
Preview at http://boegel-hod.readthedocs.org/en/cli_docs/Command_line_interface.html


boegel:
currently, this is an outlook to the CLI we want to converge too, it doesn't reflect the current state of the CLI in 3.0.0dev.

ehiggs:
I don't think `hod create [cluster label]` is terribly useful. Users seem ok with using jobids for qusb qdel, etc. So they could use them for other work, imo.

boegel:
They idea is to let `hod` be the only-and-only interface that users should worry about. To me, this includes not exposing users to job IDs.

I guess we could make labels optional, and use the job ID as a default is no label was specified.

I agree this is not a blocker though, for now we can move on using job IDs as labels (hardcoded); we can revisit this later.

boegel:
Jenkins: test this please

ehiggs:
jenkins: test this please

ehiggs:
:+1: 

Thanks!

PR #44, opened 2015-08-13
further redesign of CLI

boegel:
works further towards what is outlined in #43

### DIFF
--- a/bin/hod
+++ b/bin/hod
@@ -1,55 +1,48 @@
-#!/usr/bin/env python
+#!/bin/bash
+# #
+# Copyright 2009-2015 Ghent University
+#
+# This file is part of hanythingondemand
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
+# and the Department of Economy, Science and Innovation (EWI) (http://www.ewi-vlaanderen.be/en).
+#
+# http://github.com/hpcugent/hanythingondemand
+#
+# hanythingondemand is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation v2.
+#
+# hanythingondemand is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
 
-import sys
+# Python 2.6 or more recent 2.x required
+REQ_MAJ_PYVER=2
+REQ_MIN_PYVER=6
+REQ_PYVER=${REQ_MAJ_PYVER}.${REQ_MIN_PYVER}
 
-try:
-    from mpi4py import MPI
-except ImportError, e:
-    sys.stderr.write('Error: Could not import mpi4py. Is there an environment module providing MPI support loaded?\n')
-    sys.stderr.write('Underlying error is: "%s"\n' % e)
-    sys.exit(1)
+# make sure Python version being used is compatible
+pyver=`python -V 2>&1 | cut -f2 -d' '`
+pyver_maj=`echo $pyver | cut -f1 -d'.'`
+pyver_min=`echo $pyver | cut -f2 -d'.'`
 
-from hod import VERSION
-from hod.subcommands import create, listcmd, genconfig, helptemplate, dists
+if [ $pyver_maj -ne $REQ_MAJ_PYVER ]
+then
+    echo "ERROR: EasyBuild is currently only compatible with Python v${REQ_MAJ_PYVER}.x, found v${pyver}" 1>&2
+    exit 1
+fi
+if [ $pyver_min -lt $REQ_MIN_PYVER ]
+then
+    echo "ERROR: EasyBuild requires Python v${REQ_PYVER} or a more recent v${REQ_MAJ_PYVER}.x, found v${pyver}." 1>&2
+    exit 2
+fi
 
-SUBCOMMANDS = [
-    create.CreateSubCommand,
-    listcmd.ListSubCommand,
-    dists.DistsSubCommand,
-    helptemplate.HelpTemplateSubCommand,
-    genconfig.GenConfigSubCommand,
-]
-
-CMDS = dict([(sc.CMD, sc) for sc in SUBCOMMANDS])
-
-
-def usage(fh=None):
-    """Print the usage information for 'hod'."""
-    if fh is None:
-        fh = sys.stdout
-
-    fh.write("hod: hanythingondemand - Run services within a torque cluster\n")
-    fh.write("usage: hod [--version] [--help] <subcommand> <subcommand-args>\n")
-    fh.write("Available subcommands:\n")
-    for sc in SUBCOMMANDS:
-        fh.write('    {0:16}{1}\n'.format(sc.CMD, sc.HELP))
-
-def main(args):
-    if len(args) == 1:
-        sys.exit(usage())
-    elif args[1] in ['--version', '-V']:
-        print VERSION
-        sys.exit(0)
-    elif args[1] in ['--help', '-h', '-H']:
-        usage()
-        sys.exit(0)
-    if args[1] not in CMDS:
-        sys.stderr.write("ERROR: Unknown subcommand '%s'\n")
-        usage(fh=sys.stderr)
-        sys.exit(1)
-
-    cmd = CMDS[args[1]]()
-    cmd.run(args[1:])
-
-if __name__ == '__main__':
-    sys.exit(main(sys.argv))
+python -O -m hod.main $@
--- a/bin/hod
+++ b/bin/hod
@@ -24,9 +24,9 @@
 # along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
 # #
 
-# Python 2.6 or more recent 2.x required
+# Python 2.7 or more recent 2.x required
 REQ_MAJ_PYVER=2
-REQ_MIN_PYVER=6
+REQ_MIN_PYVER=7
 REQ_PYVER=${REQ_MAJ_PYVER}.${REQ_MIN_PYVER}
 
 # make sure Python version being used is compatible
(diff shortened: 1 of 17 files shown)


boegel:
@ehiggs: please review

bin/hod
```diff
@@ -1,55 +1,48 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+
+# Python 2.6 or more recent 2.x required
+REQ_MAJ_PYVER=2
+REQ_MIN_PYVER=6
```
ehiggs: 7

bin/hod
```diff
@@ -1,55 +1,48 @@
...
+REQ_PYVER=${REQ_MAJ_PYVER}.${REQ_MIN_PYVER}
+
+# make sure Python version being used is compatible
+pyver=`python -V 2>&1 | cut -f2 -d' '`
+pyver_maj=`echo $pyver | cut -f1 -d'.'`
+pyver_min=`echo $pyver | cut -f2 -d'.'`
+
+if [ $pyver_maj -ne $REQ_MAJ_PYVER ]
+then
+    echo "ERROR: EasyBuild is currently only compatible with Python v${REQ_MAJ_PYVER}.x, found v${pyver}" 1>&2
```
ehiggs: `s/EasyBuild/hod`

bin/hod
```diff
@@ -1,55 +1,48 @@
...
+    exit 1
+fi
+if [ $pyver_min -lt $REQ_MIN_PYVER ]
+then
+    echo "ERROR: EasyBuild requires Python v${REQ_PYVER} or a more recent v${REQ_MAJ_PYVER}.x, found v${pyver}." 1>&2
```
ehiggs: `s/EasyBuild/hod/`

hod/main.py
```diff
@@ -0,0 +1,81 @@
...
+    helptemplate.HelpTemplateSubCommand,
+    genconfig.GenConfigSubCommand,
+]
+
+SUBCOMMAND_CLASSES = dict([(sc.CMD, sc) for sc in SUBCOMMANDS])
+
+
+def usage():
+    """Print the usage information for 'hod'."""
+    usage = "hod: hanythingondemand - Run services within an HPC cluster\n"
```
ehiggs: Use triple quote and dedent.

hod/main.py
```diff
@@ -0,0 +1,81 @@
...
+    subcmd_class = None
+    for subcmd in SUBCOMMAND_CLASSES:
+        if subcmd in args:
+            subcmd_class = SUBCOMMAND_CLASSES[subcmd]
+            break
+
+    if subcmd_class:
+        subcmd_class().run(args[1:].remove(subcmd))
+
+    elif len([arg for arg in args if not arg.startswith('-')]) > 1:
```
ehiggs: This doesn't seem to support `--version`. That might be ok if we're happy to put version in the usage and just remove the flag.

hod/main.py
```diff
@@ -0,0 +1,81 @@
...
+        usage += '    {0:16}{1}\n'.format(sc.CMD, sc.HELP)
+
+    return usage
+
+
+def main(args):
+    """Parse options and run specified subcommand."""
+    # check if a subcommand is specified
+    subcmd_class = None
+    for subcmd in SUBCOMMAND_CLASSES:
```
ehiggs: Move to a function.

hod/node/node.py
```diff
@@ -23,31 +23,51 @@
...
+try:
+    import netifaces
+    import netaddr
+    # netifaces and netaddr are available, no need guard against import errors
+    def netifaces_netaddr_available(fn):
+        """No-op decorator."""
+        return fn
+
+except ImportError as err:
+    def netifaces_netaddr_available(_):
```
ehiggs: This seems really needless given that netifaces and netaddr are requirements for install.

hod/options.py
```diff
@@ -0,0 +1,55 @@
...
+
+@author: Ewan Higgs (Universiteit Gent)
+@author: Kenneth Hoste (Universiteit Gent)
+"""
+from vsc.utils.generaloption import GeneralOption
+
+from hod import VERSION as HOD_VERSION
+
+
+class HODOptions(GeneralOption):
```
ehiggs: Why make this a type? We only want to avoid copypasta of the opts dict. So just make `config_options = {...}` and then `CreateOptions` can use `opts : hod.options.config_options` instead of having to inherit.

bin/hod
boegel: OK. But why do we require 2.7 exactly? Just wondering, fine by me.

bin/hod
boegel: oops, nice catch

bin/hod
boegel: got it

hod/main.py
boegel: Hmm, you're right... Will fix.

hod/node/node.py
boegel: Well, so is `mpi4py`?

If people simply unpack a source tarball and set `$PYTHONPATH` as well, things shouldn't go down hard, but somewhat gracefully?

hod/main.py
boegel: ok

hod/options.py
boegel: I was mostly trying to avoid duplicating code, but OK, changed this to what you suggested.

boegel:
@ehiggs: remarks fixed, please rereview

hod/node/node.py
ehiggs: mpi4py is 'special' since you need to have the LD_LIBRARY_PATH pointed to the appropriate MPI libraries. Most other python packages don't have this.

bin/hod
ehiggs: It seemed to work with python 2.6 but Jenkins found something. Instead of fixing it, we can just rely on 2.7.

bin/hod
```diff
@@ -1,55 +1,48 @@
...
-        sys.stderr.write("ERROR: Unknown subcommand '%s'\n")
-        usage(fh=sys.stderr)
-        sys.exit(1)
-
-    cmd = CMDS[args[1]]()
-    cmd.run(args[1:])
-
-if __name__ == '__main__':
-    sys.exit(main(sys.argv))
+#!/bin/bash
```
ehiggs: There's not much point in this script. We can just run import platform and check the [major minor versions within python](https://docs.python.org/2/library/platform.html#platform.python_version_tuple).
```
In [6]: platform.python_version_tuple()
Out[6]: ('2', '7', '10')
```

hod/main.py
```diff
@@ -0,0 +1,85 @@
...
+    return None, args
+
+
+def main(args):
+    """Parse options and run specified subcommand."""
+    subcmd, args = init_subcmd(args)
+    if subcmd:
+        subcmd.run(args)
+
+    elif len([arg for arg in args if not arg.startswith('-')]) > 1:
```
ehiggs: If there are arguments without '-' prefixes, we print usage here. Then the 'else' part doesn't actually check the '-' prefix arguments. Aren't we supposed to check for `--version` or `--help`?

hod/options.py
```diff
@@ -0,0 +1,57 @@
...
+    'dist': ("Prepackaged Hadoop distribution (e.g.  Hadoop/2.5.0-cdh5.3.1-native). "
+             "This cannot be set if --hodconf is set", 'string', 'store', None),
+    'workdir': ("Working directory", 'string', 'store', None),
+}
+
+
+_log = fancylogger.getLogger('create', fname=False)
+
+
+def validate_pbs_option(options):
```
ehiggs: +1 putting this here. There are others we can move here too like `hod.config.config.resolve_dist_path` and friends

hod/main.py
```diff
@@ -0,0 +1,85 @@
...
+SUBCOMMAND_CLASSES = dict([(sc.CMD, sc) for sc in SUBCOMMANDS])
+
+
+def usage():
+    """Print the usage information for 'hod'."""
+    usage = "hanythingondemand version %s - Run services within an HPC cluster\n" % VERSION
+    usage += "usage: hod <subcommand> [subcommand options]\n"
+    usage += "Available subcommands (one of these must be specified!):\n"
+    for sc in SUBCOMMANDS:
+        usage += '    {0:16}{1}\n'.format(sc.CMD, sc.HELP)
```
ehiggs: We're not using EXAMPLE here; it doesn't seem it's used anywhere.

hod/subcommands/helptemplate.py
```diff
@@ -57,7 +57,7 @@ def mk_fmt_str(fields, resolver):
 class HelpTemplateSubCommand(SubCommand):
     """Implementation of 'help-template' subcommand."""
     CMD = 'help-template'
-    EXAMPLE = "--config=<hod.conf file> --workdir=<working directory>"
+    EXAMPLE = ''
```
ehiggs: If it's None then we can more easily neglect printing it (and the associated newline).

ehiggs:
Looks like a good improvement. I'm not convinced by the decorator for netaddr and netifaces. mpi4py and pbs python are a bit different since they need non python installed components available at runtime. 

I'm also not sure we need the script to wrap it since all the stuff there can be done from within python. It might be ok to check, for example, the version. And we can check that they can import the nastier glue modules (mpi4py and pbs python) But if we get into adversarial environment detection, we will always fall short. Users can always mangle their environment in creative ways. So I think we should assume they can sort out the normal modules like netaddr and netifaces.

hod/node/node.py
boegel: My main reason for this approach is to make the modules that are not included in the standard library only required when they are really needed, i.e. when you're actually triggering the functionality where they're required (without stepping away from top-level imports).

In my development environment, where I don't have `pbs_python`, `mpi4py`, `netifaces`, etc., I want to be able to run simple things like `hod` and `hod create --help` without running into import errors. If I do call out to the functionality where these are needed, I want a clean error message, not a traceback (which looks like a bug in HOD, while it's actually just a missing module).

Whether these modules include extra libraries as well matters little imho.

I don't see the harm in these decorators; if anything, they indicate which functionality depends on modules that are not in the Python stlib.

Even with the bootstrap, EasyBuild is still installed via `easy_install`, so your last sentence makes no sense. :-)

bin/hod
boegel: What if `hod/main.py` fails to parse in Python 3, or Python 2.x with x < 7?

Also, you should check `sys.version` I think (although I'm not sure if `platform.python_version_tuple` is different).

hod/options.py
boegel: Sure, we can move other stuff here. But not in this PR. :-)

hod/main.py
boegel: The intention of this is to print usage whenever no known subcommand is used.

Checking for `--help` wouldn't make any difference, as we'd still print usage then (like we do in the else bit). And the version is included in the usage now, so there's no use for a top-level `--version`.

hod/main.py
boegel: It's used in the `usage()` of `SubCommand`; printing usage info per subcommand is beyond scope for the top-level usage?

hod/subcommands/helptemplate.py
boegel: we're already using `if EXAMPLE` in `SubCommand.usage()`, but fine, I'll change this to a `None`

bin/hod
ehiggs: `platform.python_version_tuple` parses `sys.version`: 

```python
def python_version_tuple():

    """ Returns the Python version as tuple (major, minor, patchlevel)
        of strings.

        Note that unlike the Python sys.version, the returned value
        will always include the patchlevel (it defaults to 0).

    """
    return tuple(string.split(_sys_version()[1], '.'))
```

hod/main.py
ehiggs: ok. :+1:

bin/hod
boegel: OK, fine, but it's still safer to check the version before actually running `hod/main.py`.

Once we have both py2/py3 compatibility (some day...), it may be a different story. Until then, doing the check in `hod/main.py` is a risk of running into `SyntaxError`s imho.

ehiggs: I guess if someone runs `module load hanythingondemand` and then `module load Python` which might default to Python 3, and they ignore the error about the new Python version conflicting with the one needed by hanythingondemand, then yes, they can get a syntax error. So if you feel it's required, we can do this using the script.

ehiggs:
Jenkins, test this please

Issue #36, opened 2015-08-11
HOD only works when it's installed as a module


boegel:
Currently, HOD requires that either:

* `$EBMODNAMEHANYTHINGONDEMAND` is defined with the module name under which HOD is available
* or, that the `hanythingondemand` module was loaded with Tcl/C environment modules, i.e., that the `hanythingondemand` module name is listed in `$LOADEDMODULES`

This is rather clunky, and could probably be relaxed a bit. The only thing that is really needed is figuring out how `$PYTHONPATH` should be set, and which dependency modules should be loaded.

ehiggs:
We kinda baked this in with `--hod-module`. But at least it's now specified rather than detected. If we're happy with `--hod-module` I think we can close this since we've implemented something that appears to be diametrically opposed.

boegel:
The code that this issue refer to is removed in the PR for `--hod-module` (see #70), so let's close this since it's basically a duplicate now

PR #48, opened 2015-08-17
IPython notebook w/ autogen part.

ehiggs:


ehiggs:
This needs docs and more testing. And a bump to use IPython 3.2.0 or 4.0.0.

Useful docs for ipython:
http://ipython.org/ipython-doc/stable/notebook/security.html
http://stackoverflow.com/questions/29037211/how-do-i-add-a-kernel-on-a-remote-machine-in-ipython-jupyter-notebook
https://github.com/ipython/ipython/wiki/Cookbook:-Connecting-to-a-remote-kernel-via-ssh

etc/hod/IPython-notebook-3.0.0/start-notebook.sh
```diff
@@ -0,0 +1,32 @@
...
+
+cat <<EOF > ${config_dir}/ipython_notebook_config.py
+c = get_config()
+
+# Kernel config
+c.IPKernelApp.pylab = 'inline'  # if you want plotting support always
+
+# Notebook config
+c.NotebookApp.open_browser = False
+#c.NotebookApp.password = u'sha1:bcd259ccf...[your hashed password here]'
```
boegel: hmm, what's the deal with this? uncommenting makes the notebook available without requiring a password?
if that's right, please add a comment above this line stating that

hod/config/autogen/ipython_notebook.py
```diff
@@ -0,0 +1,71 @@
...
+# Notebook config
+c.NotebookApp.ip = '$ip'
+c.NotebookApp.open_browser = False
+c.NotebookApp.password = u'sha1:{hashed_password}'
+# It is a good idea to put it on a known, fixed port
+c.NotebookApp.port = 9999
+"""
+    template = string.Template(template)
+    file_contents = template.substitute(
+            hashed_password='banana',
```
boegel: use `getpass` rather than hardcoding this super-safe passwd?

(not sure if that's really an option)

boegel:
Should be updated to at least IPython 3.2.1 and Python 2.7.10 (not sure if we should go with IPython 4.0.0 and Jupyter 1.0.0 yet, since it may be too much of a black hole w.r.t. time).

hod/config/autogen/ipython_notebook.py
ehiggs: It's a good idea. But I think this won't work since this will be run headlessly.

This could possibly be integrated with `hod connect` somehow.

boegel: Maybe by picking up an environment variable specifying a password (or hash?), and then passing that to `ipython_notebook_config`?

etc/hod/IPython-notebook-3.2.1/ipython.conf
```diff
@@ -0,0 +1,13 @@
+#-*- cfg -*-
+# vim: ft=cfg
+[Unit]
+Name=ipython
+RunsOn=master
+
+[Service]
+ExecStart=nohup $$EBROOTHANYTHINGONDEMAND/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
```
boegel: what if `hanythingondemand` is not provided via a module?

this is another instance of #36 (actually, it's worse, since there's no fallback env var like `EBMODNAMEHANYTHINGONDEMAND`)

boegel: no need to fix this here, just taking note of it basically...

boegel:
Tried to get an HOD cluster running with this.

I see `ipython notebook` running (only) on the head node, but also see this popping up in `hod.local` (as last debug log message):

```
2015-08-19 18:38:44,559 DEBUG	   runpy.Command   mpi: 0 MainThread  Timeout occured with cmd nohup $EBROOTHANYTHINGONDEMAND/etc/hod/IPython-notebook-3.2.1/start-notebook.sh. took more than 120 secs to complete.
```

Is this something to worry about?

ehiggs:
That's a message we want to avoid. It is there because `hod.command.Command` spawns a subprocess and checks if it returned flow of control back to hod in 120 seconds. If it didn't, it considers it an error. So here, it might be better for our IPython script to fork like a daemon.

e.g. from [how yarn used to do it](https://github.com/apache/hadoop/blob/release-2.5.1/hadoop-yarn-project/hadoop-yarn/bin/yarn-daemon.sh#L124):

```
    nohup nice -n $YARN_NICENESS "$HADOOP_YARN_HOME"/bin/yarn --config $YARN_CONF_DIR $command "$@" > "$log" 2>&1 < /dev/null &
```

They've since gone full :stuck_out_tongue_winking_eye: and written some monstrosity that hides the implementation behind hundreds (thousands?) of lines of bash littered around their repos.

boegel:
@ehiggs: let's open an issue on that? I'll look into getting connected to IPython, but it seems like this PR is good to go now. Although I'd like to discuss the password aspect in here....

hod/config/autogen/ipython_notebook.py
boegel: or how about generating a random string and using that as a password? the password or hash doesn't need to be available to the user, or does it?

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+module load IPython/3.2.1-intel-2015a-Python-2.7.10
```
boegel: should we include `set -e` in here?

hod/config/autogen/ipython_notebook.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Kernel config
+c.IPKernelApp.pylab = 'inline'  # if you want plotting support always
+
+# Notebook config
+c.NotebookApp.ip = '$ip'
+c.NotebookApp.open_browser = False
+c.NotebookApp.password = u'sha1:{hashed_password}'
+# It is a good idea to put it on a known, fixed port
+c.NotebookApp.port = 9999
```
boegel: fix this to 8888, the default port for IPython, to avoid surprises/confusion?

hod/config/autogen/ipython_notebook.py
```diff
@@ -0,0 +1,71 @@
...
+##
+"""
+IPython Notebook autoconfiguration.
+
+@author: Ewan Higgs (Ghent University)
+"""
+
+import string
+
+def ipython_notebook_config(workdir, node_info):
```
boegel: `workdir` is unused, the config should set this as the default 'home' directory for the notebook?

if `node_info` is used, use `_` instead

hod/config/autogen/ipython_notebook.py
```diff
@@ -0,0 +1,71 @@
...
+    Generate config 
+    '''
+    template = r"""
+c = get_config()
+
+# Kernel config
+c.IPKernelApp.pylab = 'inline'  # if you want plotting support always
+
+# Notebook config
+c.NotebookApp.ip = '$ip'
```
boegel:
```
KernelManager.ip : Unicode
Default: u’‘

Set the kernel’s IP address [default localhost]. If the IP address is something other than localhost, then Consoles on other machines will be able to connect to the Kernel, so be careful!
```

do we really want/need to specify the IP here?

hod/config/autogen/ipython_notebook.py
boegel: `NotebookApp.ipython_dir` and `NotebookApp.notebook_dir` should be set to `workdir`,

hod/config/autogen/ipython_notebook.py
boegel: NotebookApp.ipython_dir and NotebookApp.notebook_dir should be set to workdir,

hod/config/autogen/ipython_notebook.py
boegel: this probably doesn't work because `banana` is not a proper hash?

ehiggs: Yeah I think if we see that a user will be using an autogen ipython thingy they will need a password at job submission time. Then we can use `getpass`.

boegel:
I've been playing around with this, seems to work, if issues pop up we can tackle them in a follow-up PR.

PR #50, opened 2015-08-19
Improve import fail error msg

boegel:
error message like `ImportError: No module named PBSQuery (No module named PBSQuery)` may be too cryptic for users?

Now results in this:

```
ImportError: No module named PBSQuery; pbs_python is not available (No module named PBSQuery; pbs_python is not available)
```


boegel:
Jenkins: test this please

PR #51, opened 2015-08-19
check whether resolved config file path exists

boegel:
Error before:

```
IOError:  ([Errno 2] No such file or directory: '/vscmnt/gent_vulpix/_/user/home/gent/vsc400/vsc40023/hanythingondemand/etc/hod/IPython-notebook-3.2.1/hod.conf')
```

error now:

```
ValueError: Config file for specified dist 'IPython-notebook-3.2.1' does not exist: /vscmnt/gent_vulpix/_/user/home/gent/vsc400/vsc40023/hanythingondemand/etc/hod/IPython-notebook-3.2.1/hod.conf (Config file for specified dist 'IPython-notebook-3.2.1' does not exist: /vscmnt/gent_vulpix/_/user/home/gent/vsc400/vsc40023/hanythingondemand/etc/hod/IPython-notebook-3.2.1/hod.conf)
```


hod/config/config.py
```diff
@@ -409,14 +409,21 @@ def avail_dists():
...
     1. config if it's defined.
     2. The expanded dist path if config is not defined.
-    '''
+    """
     if config:
-        return config
+        if os.path.exists(config):
+            path = config
+        else:
+            raise ValueError("Specified config file '%s' does not exist." % config)
```
boegel: hmm, this should probably be `log.eror("..."); sys.exit(1)`, to avoid throwing a traceback at the user?

boegel:
@ehiggs: seems like some tests will need fixing for this one

hod/config/config.py
ehiggs: No I think this is ok. We just need to wrap the call in a try which will catch this ValueError (and change below runtime error to ValueError) and that will report the error and exit.

Maybe we should use [IOError](https://docs.python.org/2.7/library/exceptions.html#exceptions.IOError)

boegel: `resolve_config_paths` is called in 3 places at this moment (2x in `hodproc.py`, 1x in `hodjob.py`)

Wrapping that call in a `try`/`except` would result in adding about 3x4 lines (`try`+`except`+`log.error`+`sys.exit`), while just exiting here (rather than raising an error to then exit later) would only result in 1 or 2 lines extra...

PR #52, opened 2015-08-19
fix hodjob.get_hod

boegel:
@ehiggs: `get_hod` was utterly broken, but the tests failed to notice...

Can we fix this?

`get_hod()` should always be able to find `hod-local` (the default), since that's the only way we use it


hod/rmscheduler/hodjob.py
```diff
@@ -78,28 +78,28 @@ def get_hod(self, exe_name='hod-local'):
         """Get the full path of the exe_name
              -look in bin or bin / .. / hod /
         """
-        fullscriptname = os.path.abspath(sys.argv[0])
-
-        bindir = os.path.dirname(fullscriptname)
-        hodpythondir = os.path.abspath("%s/.." % bindir)
-        hoddir = os.path.abspath("%s/../hod" % bindir)
+        fullscriptname = os.path.abspath(sys.argv[0])  # <prefix>/hod/main.py
+        topdir = os.path.dirname(os.path.dirname(fullscriptname))  # <prefix>
```
boegel: this is not going to be correct if `hanythingondemand` was installed via `easy_install`, since then you'd have `<libdir>/lib/python2.7/site-packages/hanythingondemand*egg/hod/main.py`...

boegel: for finding binaries like `hod-local`, we should go via `$PATH` unless `$HOD_BINDIR` is defined?

boegel:
forget this, `get_hod` is removed in #54 anyway

PR #55, opened 2015-08-19
Use pkg_resources to find etc dir.

### LINKED ISSUE #53: hod dist doesn't work if we start the python process from `hod.main`
hod dist doesn't work if we start the python process from `hod.main`.

`hod dist` looks at the top level executable (within the python process) to determine where the etc dir is. Since `hod.main` was moved into the hod tree and out of bin, it can no longer find the dists.

ehiggs:
We can't rely on finding the bin dir and going ../etc to find the appropriate dir for configs. thankfully pkg_resources exists so we don't have to be shonky. We can rely on pkg_resources to be shonky for us. 

Fixes #53


hod/config/config.py
```diff
@@ -391,22 +393,23 @@ def write_service_config(outfile, data_dict, config_writer, template_resolver):
...
+    distpath = resolve_dists_dir()
+    distpath = mkpath(distpath, dist, 'hod.conf')
     return distpath
 
 def avail_dists():
     """Return a list of available distributions"""
-    return os.listdir(resolve_dists_dir())
+
+    pkg = Requirement.parse("hanythingondemand")
+    return resource_listdir(pkg, os.path.join('etc', 'hod'))
```
boegel: this is almost copy-pasting, so this calls for some refactoring?

maybe just use `os.listdir` here?

if the `etc` files are wrapped in a `.egg` or whatever we're fucking anyways?

boegel: or you need something like `get_dist_file('hod.conf')` too, which uses `resource_string`

boegel:
@ehiggs: please add a test for `avail_dists` too

hod/config/config.py
ehiggs: Good points.

I think `os.listdir` and `resource_listdir` are basically the same but `resource_listdir` will work on a zipped egg. So if we ever want to distribute with zipped eggs, it's less wrong now. Other code will break, but at least this bit is correct.

I think a refactor of the rest of the code to work with zipped eggs is a good idea but not in scope for this PR.

ehiggs:
Added test for avail_dists.

hod/config/config.py
```diff
@@ -34,10 +34,12 @@
 from copy import deepcopy
 from importlib import import_module
 from os.path import join as mkpath, dirname, realpath
+from pkg_resources import Requirement, resource_filename, resource_listdir
```
boegel: we need to include `setuptools` as a dep in `setup.py` because of this, no?

hod/config/config.py
boegel: OK, so let's just open an issue for this, something like "access dist files via pkg_resources.resource_string in case they're in a zipped egg"

hod/config/config.py
ehiggs: [`setup.py` uses `setuptools` to install the software](https://github.com/hpcugent/hanythingondemand/blob/master/setup.py#L34).

boegel: OK, but that's just for installation. Now `setuptools` is also a runtime dep.

boegel:
:+1: :facepunch:

PR #56, opened 2015-08-20
Add tests to run through main and local.

ehiggs:
Local doesn't really assert anything since it's just dispatching. Main flexes
the command parsing code.

### DIFF
--- a/test/unit/test_hod_local.py
+++ b/test/unit/test_hod_local.py
@@ -34,15 +34,12 @@ import hod.mpiservice as hm
 
 class TestHodLocal(unittest.TestCase):
     def test_local_no_args(self):
-        with patch('sys.exit'):
-            hl.main([])
+        self.assertRaises(SystemExit, hl.main, [])
 
     def test_master_rank(self):
-        with patch('sys.exit'):
-            with patch('mpi4py.MPI.COMM_WORLD', Mock(rank=hm.MASTERRANK)):
-                hl.main([])
+        with patch('mpi4py.MPI.COMM_WORLD', Mock(rank=hm.MASTERRANK)):
+            self.assertRaises(SystemExit, hl.main, [])
 
     def test_slave_rank(self):
-        with patch('sys.exit'):
-            with patch('mpi4py.MPI.COMM_WORLD', Mock(rank=hm.MASTERRANK + 1)):
-                hl.main([])
+        with patch('mpi4py.MPI.COMM_WORLD', Mock(rank=hm.MASTERRANK + 1)):
+            self.assertRaises(SystemExit, hl.main, [])
--- a/test/unit/test_hod_local.py
+++ b/test/unit/test_hod_local.py
@@ -1,0 +1,48 @@
+###
+# Copyright 2009-2014 Ghent University
+#
+# This file is part of hanythingondemand
+# originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
+# with support of Ghent University (http://ugent.be/hpc),
+# the Flemish Supercomputer Centre (VSC) (https://vscentrum.be/nl/en),
+# the Hercules foundation (http://www.herculesstichting.be/in_English)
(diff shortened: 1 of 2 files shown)


test/unit/test_hod_main.py
```diff
@@ -0,0 +1,73 @@
...
+import hod.subcommands.dists
+
+class TestHodMain(unittest.TestCase):
+    def test_main_no_args(self):
+        with patch('sys.exit'):
+            hm.main([])
+
+    def test_main_bad_cmd(self):
+        with patch('sys.exit'):
+            hm.main(['hod', 'banana'])
```
boegel: use `self.assertRaises`?

test/unit/test_hod_main.py
```diff
@@ -0,0 +1,73 @@
...
+'''
+
+import unittest
+
+from mock import patch
+
+import hod.main as hm
+import hod.subcommands.dists
+
+class TestHodMain(unittest.TestCase):
```
boegel: use `EnhancedTestCase` from `vsc.utils.testing`, has some useful extra stuff like `self.assertErrorRegex`

test/unit/test_hod_main.py
```diff
@@ -0,0 +1,73 @@
...
+        with patch('sys.exit'):
+            hm.main([])
+
+    def test_main_bad_cmd(self):
+        with patch('sys.exit'):
+            hm.main(['hod', 'banana'])
+
+    def test_main_good_cmd(self):
+        with patch('sys.exit'):
+            hm.main(['hod', 'dists'])
```
boegel: no need to patch `sys.exit` here?

test/unit/test_hod_main.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_main_help(self):
+        with patch('sys.exit'):
+            hm.main(['hod', '--help'])
```
boegel: no need to patch `sys.exit`?

ehiggs: main actually prints usage and then calls exit(0). Which I'm ok with. Just means I need to patch it out here.

test/unit/test_hod_main.py
ehiggs: It doesn't raise. It hard exists. Which is fine, imo.

boegel: `sys.exit` is basically equivalent to `raise SystemExit`. So it does raise.

test/unit/test_hod_main.py
boegel: `self.assertErrorRegex(SystemExit, 0, hm.main, ['hod', '--help'])` is a better test imho (since you're checking the exit code too)

test/unit/test_hod_main.py
```diff
@@ -0,0 +1,67 @@
...
+
+import hod.main as hm
+import hod.subcommands.dists
+
+class TestHodMain(unittest.TestCase):
+    def test_main_no_args(self):
+        self.assertRaises(SystemExit, hm.main, [])
+
+    def test_main_bad_cmd(self):
+        self.assertRaises(SystemExit, hm.main, ['hod', 'banana'])
```
boegel: bad command, so should be exit code != 0, right? check that

test/unit/test_hod_local.py
```diff
@@ -0,0 +1,45 @@
...
+import hod.local as hl
+import hod.mpiservice as hm
+
+class TestHodLocal(unittest.TestCase):
+    def test_local_no_args(self):
+        self.assertRaisesRegexp(SystemExit, '1', hl.main, [])
+
+    def test_master_rank(self):
+        with patch('mpi4py.MPI.COMM_WORLD', Mock(rank=hm.MASTERRANK)):
+            self.assertRaisesRegexp(SystemExit, '1', hl.main, [])
```
boegel: why exit 1 here?

ehiggs: It tries to run the code but runs into some other problem like no `--workdir` defined or whatever.

PR #58, opened 2015-08-21
We don't fully support zipped eggs yet since we do 'open' on the config files

ehiggs:
This is a small fix to installation.


setup.py
```diff
@@ -96,6 +96,7 @@ def find_files(*dirs):
     'scripts': ['bin/hod'],
     'cmdclass' : {'test': TestCommand},
     'long_description': open(os.path.join(os.path.dirname(__file__), 'README.md')).read(),
+    'zip_safe': False
```
boegel: add `,`

PR #59, opened 2015-08-21
Add the hod.conf to PATH so we can run scripts; also fix start-notebook.

ehiggs:
…ok.sh

Also, add an argument to start-notebook to give it the localworkdir/conf and we put
the ipython_notebook config in there.

### DIFF
--- a/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
+++ b/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
@@ -14,8 +14,11 @@ if [[ ! -d "$1" ]] ; then
     exit 1
 fi
 
-config_dir="$1/ipython_notebook"
+config_dir="$1/conf/ipython_notebook"
+log_dir="$1/log"
+
 mkdir -p "$config_dir"
+mkdir -p "$log_dir"
 
 cat <<EOF > "${config_dir}/ipython_notebook_config.py"
 c = get_config()
@@ -41,4 +44,4 @@ export PYTHONPATH=$PYTHONPATH:$EBROOTPYTHON/lib/python2.7/site-packages/
 export PYSPARK_DRIVER_PYTHON=ipython 
 export PYSPARK_DRIVER_PYTHON_OPTS="notebook --profile-dir=${config_dir}" 
 export PYSPARK_SUBMIT_ARGS="--master yarn --deploy-mode client"
-pyspark --master yarn
+pyspark --master yarn >"$log_dir/pyspark.stdout" 2>"$log_dir/pyspark.stderr"
--- a/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
+++ b/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
@@ -32,4 +32,7 @@ export PYTHONPATH="$EBROOTSPARK/python/:$EBROOTSPARK/python/lib/py4j-0.8.2.1-src
 
 export PYTHONPATH=$PYTHONPATH:$EBROOTPYTHON/lib/python2.7/site-packages/ 
 
-ipython notebook --profile-dir="${config_dir}"
+#ipython notebook --profile-dir="${config_dir}"
+export PYSPARK_DRIVER_PYTHON=ipython 
+export PYSPARK_DRIVER_PYTHON_OPTS="notebook --profile-dir=${config_dir}" 
+pyspark
--- a/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
+++ b/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
@@ -10,7 +10,7 @@ module load Spark/1.4.1
 ipython profile create nbserver
 
 config_dir="$1/ipython_notebook"
-mkdir "$config_dir"
+mkdir -p "$config_dir"
 
 cat <<EOF > "${config_dir}/ipython_notebook_config.py"
 c = get_config()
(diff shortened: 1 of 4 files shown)


hod/hodproc.py
```diff
@@ -115,6 +115,9 @@ def distribution(self, *master_template_args, **kwargs):
         _setup_config_paths(m_config, resolver)
 
         master_env = dict([(v, os.getenv(v)) for v in m_config.master_env])
+        # There may be scripts in the hod.conf dir so add it to the PATH
+        master_env['PATH'] = master_env.get('PATH', os.getenv('PATH')) + ':' + m_config.hodconfdir
```
boegel: use `os.pathsep` rather than `:`

hod/config/config.py
```diff
@@ -260,7 +267,9 @@ def invalid_fields(obj):
     """Return list of fields which are empty."""
     bad_fields = []
     for attr in obj.__slots__:
-        if attr not in obj.OPTIONAL_FIELDS and not getattr(obj, attr):
+        if (attr not in obj.OPTIONAL_FIELDS
+                and not getattr(obj, attr)
+                and not attr.startswith('_')):
```
boegel:
```python
if not (attr in obj.OPTIONAL_FIELDS or getattr(obj, attr) or attr.startswith('_')):
```

hod/config/config.py
```diff
@@ -179,6 +182,10 @@ def localworkdir(self):
     def configdir(self):
         return mkpath(self.localworkdir, 'conf')
 
+    @property
+    def hodconfdir(self):
+        return self._hodconfdir
```
boegel: why use a property rather than make this `self.hodcondir` directly?!

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
```diff
@@ -9,9 +9,10 @@ module load Spark/1.4.1
 
 ipython profile create nbserver
 
-config_dir=(mktmp -d ipython_notebook)
+config_dir="$1/ipython_notebook"
+mkdir "$config_dir"
```
boegel: use `mkdir -p`

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
```diff
@@ -9,9 +9,10 @@ module load Spark/1.4.1
 
 ipython profile create nbserver
 
-config_dir=(mktmp -d ipython_notebook)
+config_dir="$1/ipython_notebook"
```
boegel: add a check to make sure that `$1` is actually set?

we don't want to try and create `/ipython_notebook`...

hod/config/config.py
```diff
@@ -163,6 +165,7 @@ def _get_list(name):
 
         self.modules = _get_list('modules')
         self.master_env = _get_list('master_env')
+        self._hodconfdir = fileobj_dir
```
boegel: why here in the middle? keep `self._` ones separate?

ehiggs:
Spark 1.4.1 is now using their own system for working w/ ipython notebook so I changed the script to work with that.

I tried it on an example dataset and it's taking forever to load a file and count the lines. Knowing spark it really could take 10mins to do this on an 8mb file but really I think it's hanging. :(

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
ehiggs: :+1:

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
ehiggs: :+1:

hod/config/config.py
ehiggs: The fields here are actual fields in the config file. I now filter out the `_prefix` fields so they aren't in the config. Thus the property isn't in the actual config file.

hod/hodproc.py
ehiggs: Good catch! :+1:

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
```diff
@@ -27,8 +28,11 @@ c.NotebookApp.port = 8888
...
-export PYTHONPATH="$EBROOTSPARK/python/:$EBROOTSPARK:python/lib/py4j-0.8.2.1-src.zip:$PYTHONPATH"
+export PYTHONPATH="$EBROOTSPARK/python/:$EBROOTSPARK/python/lib/py4j-0.8.2.1-src.zip:$PYTHONPATH"
 
 export PYTHONPATH=$PYTHONPATH:$EBROOTPYTHON/lib/python2.7/site-packages/ 
 
-ipython notebook --profile-dir=${config_dir}
+#ipython notebook --profile-dir="${config_dir}"
+export PYSPARK_DRIVER_PYTHON=ipython 
+export PYSPARK_DRIVER_PYTHON_OPTS="notebook --profile-dir=${config_dir}" 
+pyspark
```
boegel: so rename the script to `start-pyspark.sh`?!

ehiggs:
The last commit sends stdout and stderr to log files. It appears to work but there is a warning/error:

```
2015-08-26 17:25:04,786 WARNING    runpy.Command   mpi: 0 MainThread  Problem occured with cmd nohup start-notebook.sh /path/to/localworkdir out , err Exitcode -15
```

boegel:
@ehiggs: please figure out what the warning is about before we merge this?

boegel:
@ehgigs: I think you have a minor extra commit waiting around for this one?

ehiggs:
I haven't found the fix for the warning yet. It appears to work fine otherwise. But I really want to fix the warning. :confounded:

boegel:
Are you any closer to figuring out the warning yet?

If it's 'harmless' as far as you can tell, I suggest we merge this and continue to make progress.

ehiggs:
Exitcode 15 is used in Spark's `ApplicationMaster` if user code threw an exception.

ehiggs:
Fixed the error message. You can merge this if you're happy w/ it now.

Issue #57, opened 2015-08-20
reinstate `hod batch`


boegel:

boegel:
One of our users tackled this himself:

```
So I made this configuration file halvade.conf which looks something like this:
#-*- cfg -*-
# vim: ft=cfg

[Unit]
Name=halvade
RunsOn=master

[Service]
ExecStart=bash /user/home/gent/vsc404/vsc40486/ccle/scripts/dd3f7a68-b611-4dfd-8b08-2eada063ce19.sh
ExecStop=

[Environment]

And add this halvade.conf to hod.conf at the end of the services line.
For this I have to make a separate map for each job with a different halvade.conf though, but it works.
```

ehiggs:
Done in #67!

PR #67, opened 2015-09-02
Batch subcommand

ehiggs:
This implements #57.

Usage example:

```
$ hod batch --dist Hadoop-2.5.0-cdh5.3.1-native --workdir=$MYSCRATCH/hod -n2 --script=$MYSCRATCH/wordcount/wordcount.sh
```

This works rejigs the `ConfigOptsParams` to hold all the params needed to make a `ConfigOpts` (aside from the template resolver). Moving between `ConfigOpts` and `ConfigOptsParams` use `staticmethod` functions instead of standalone functions.

The batch work is done by making the `script` option into a `ConfigOpts` object that has no config file and uses `start_script` (`ExecStart` in the configfile which we're not using) is the script but suffixed with `; qdel $PBS_JOBID` which basically just tears down the cluster.

### DIFF
--- a/hod/config/config.py
+++ b/hod/config/config.py
@@ -27,7 +27,6 @@
 """
 
 import os
-import sys
 
 from ConfigParser import NoOptionError, NoSectionError, SafeConfigParser
 from collections import Mapping, namedtuple
@@ -127,7 +126,7 @@ def parse_comma_delim_list(s):
     with no spaces on the end or beginning.
     Blanks are also removed. e.g. 'a,,b' results in ['a', 'b']
     '''
-    return [x.strip() for x in filter(lambda x: x.strip(), s.split(','))]
+    return [tok.strip() for tok in [el for el in s.split(',') if el.strip()]]
 
 
 class PreServiceConfigOpts(object):
@@ -141,7 +140,17 @@ class PreServiceConfigOpts(object):
     __slots__ = ['version', 'workdir', 'config_writer', 'directories',
                  'autogen', 'modules', 'service_configs', 'service_files', 'master_env']
 
-    OPTIONAL_FIELDS=['master_env', 'modules', 'service_configs', 'directories', 'autogen']
+    OPTIONAL_FIELDS = ['master_env', 'modules', 'service_configs', 'directories', 'autogen']
+
+    @staticmethod
+    def from_file_list(filenames, **kwargs):
+        """Create and merge PreServiceConfigOpts from a list of filenames."""
+        precfgs = [PreServiceConfigOpts(open(f, 'r'), **kwargs) for f in filenames]
+        precfg = reduce(merge, precfgs)
+        bad_fields = invalid_fields(precfg)
+        if bad_fields:
+            raise RuntimeError("A valid configuration could not be generated from the files: %s: missing fields: %s" % (filenames, bad_fields))
+        return precfg
 
     def __init__(self, fileobj, **kwargs):
         _config = load_service_config(fileobj)
@@ -210,16 +219,6 @@ class PreServiceConfigOpts(object):
                         self.config_writer, self.service_configs)
 
 
-def preserviceconfigopts_from_file_list(filenames, **kwargs):
-    """Create and merge PreServiceConfigOpts from a list of filenames."""
-    precfgs = [PreServiceConfigOpts(open(f, 'r'), **kwargs) for f in filenames]
-    precfg = reduce(merge, precfgs)
-    bad_fields = invalid_fields(precfg)
-    if bad_fields:
-        raise RuntimeError("A valid configuration could not be generated from the files: %s: missing fields: %s" % (filenames, bad_fields))
-    return precfg
-
-
 def merge(lhs, rhs):
     """
     Merge two objects of the same type based on their __slot__ list. This
@@ -227,7 +226,7 @@ def merge(lhs, rhs):
     Rules:
         List types are concatenated.
         Dict types are merged using a deep merge.
-        String types are overwritten. 
+        String types are overwritten.
     """
     if type(lhs) != type(rhs):
         raise RuntimeError('merge can only use two of the same type')
@@ -320,16 +319,16 @@ class ConfigOpts(object):
 
     def to_params(self, workdir, modules, master_template_args):
         """Create a ConfigOptsParams object from the ConfigOpts"""
-        return ConfigOptsParams(self.name, self.runs_on, self.pre_start_script,
-                self.start_script, self.stop_script, self._env, workdir,
+        return ConfigOptsParams(self.name, self._runs_on, self._pre_start_script,
+                self._start_script, self._stop_script, self._env, workdir,
                 modules, master_template_args)
 
     @staticmethod
-    def from_params(configopts, template_resolver):
+    def from_params(params, template_resolver):
         """Create a ConfigOpts from a ConfigOptsParams"""
-        return ConfigOpts(configopts.name, configopts.runs_on,
-                configopts.pre_start_script, configopts.start_script,
-                configopts.stop_script, configopts.env, template_resolver)
+        return ConfigOpts(params.name, params.runs_on,
+                params.pre_start_script, params.start_script,
+                params.stop_script, params.env, template_resolver)
 
     def __init__(self, name, runs_on, pre_start_script, start_script,
             stop_script, env, template_resolver):
@@ -401,16 +400,16 @@ class ConfigOpts(object):
 # Parameters to send over the network to allow slaves to construct hod.config.ConfigOpts
 # objects
 ConfigOptsParams = namedtuple('ConfigOptsParams', [
-    'name', 
+    'name',
     'runs_on',
-    'pre_start_script', 
+    'pre_start_script',
     'start_script',
-    'stop_script', 
-    'env', 
-    'workdir', 
-    'modules', 
+    'stop_script',
+    'env',
+    'workdir',
+    'modules',
     'master_template_kwargs'
-    ])
+])
 
 def autogen_fn(name):
     """
@@ -499,3 +498,4 @@ def resolve_config_paths(config, dist):
         raise RuntimeError('A config or a dist must be provided')
 
     return path
+
--- a/hod/config/config.py
+++ b/hod/config/config.py
@@ -305,33 +305,29 @@ class ConfigOpts(object):
     @staticmethod
     def from_file(fileobj, template_resolver):
         """Load a ConfigOpts from a configuration file."""
-        _config = load_service_config(fileobj)
+        config = load_service_config(fileobj)
 
-        name = _cfgget(_config, _UNIT_SECTION, 'Name')
-        runs_on = _parse_runs_on(_cfgget(_config, _UNIT_SECTION, 'RunsOn'))
-        pre_start_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStartPre', '')
-        start_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStart')
-        stop_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStop')
-        env = dict([(k, v) for k, v in _config.items(_ENVIRONMENT_SECTION)])
+        name = _cfgget(config, _UNIT_SECTION, 'Name')
+        runs_on = _parse_runs_on(_cfgget(config, _UNIT_SECTION, 'RunsOn'))
+        pre_start_script = _cfgget(config, _SERVICE_SECTION, 'ExecStartPre', '')
+        start_script = _cfgget(config, _SERVICE_SECTION, 'ExecStart')
+        stop_script = _cfgget(config, _SERVICE_SECTION, 'ExecStop')
+        env = dict(config.items(_ENVIRONMENT_SECTION))
 
-        return ConfigOpts(name, runs_on, pre_start_script, start_script,
-                stop_script, env, template_resolver)
+        return ConfigOpts(name, runs_on, pre_start_script, start_script, stop_script, env, template_resolver)
 
     def to_params(self, workdir, modules, master_template_args):
-        """Create a ConfigOptsParams object from the ConfigOpts"""
-        return ConfigOptsParams(self.name, self._runs_on, self._pre_start_script,
-                self._start_script, self._stop_script, self._env, workdir,
-                modules, master_template_args)
+        """Create a ConfigOptsParams object from the ConfigOpts instance"""
+        return ConfigOptsParams(self.name, self._runs_on, self._pre_start_script, self._start_script, 
+                                self._stop_script, self._env, workdir, modules, master_template_args)
 
     @staticmethod
     def from_params(params, template_resolver):
-        """Create a ConfigOpts from a ConfigOptsParams"""
-        return ConfigOpts(params.name, params.runs_on,
-                params.pre_start_script, params.start_script,
-                params.stop_script, params.env, template_resolver)
+        """Create a ConfigOpts instance from a ConfigOptsParams instance"""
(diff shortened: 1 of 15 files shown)


boegel:
Jenkins: test this please

ehiggs:
This depends on #43 having batch documentation.

ehiggs:
#43 has batch documentation now but it hasn't been merged. Shall we proceed with this?

Thanks!

boegel:
docs part is indeed covered in #43 

I'll take a close look at this one...

hod/config/config.py
```diff
@@ -135,11 +134,23 @@ class PreServiceConfigOpts(object):
...
+    OPTIONAL_FIELDS = ['master_env', 'modules', 'service_configs', 'directories', 'autogen']
+
+    @staticmethod
+    def from_file_list(filenames, **kwargs):
+        """Create and merge PreServiceConfigOpts from a list of filenames."""
+        precfgs = [PreServiceConfigOpts(open(f, 'r'), **kwargs) for f in filenames]
+        precfg = reduce(merge, precfgs)
+        bad_fields = invalid_fields(precfg)
+        if bad_fields:
+            raise RuntimeError("A valid configuration could not be generated from the files: %s: missing fields: %s" % (filenames, bad_fields))
```
boegel: line too long?

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+
+        name = _cfgget(_config, _UNIT_SECTION, 'Name')
+        runs_on = _parse_runs_on(_cfgget(_config, _UNIT_SECTION, 'RunsOn'))
+        pre_start_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStartPre', '')
+        start_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStart')
+        stop_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStop')
+        env = dict([(k, v) for k, v in _config.items(_ENVIRONMENT_SECTION)])
+
+        return ConfigOpts(name, runs_on, pre_start_script, start_script,
+                stop_script, env, template_resolver)
```
boegel: why split this line? no need imho

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
     """
-    def __init__(self, fileobj, template_resolver):
-        self._config = load_service_config(fileobj)
-        self.name = _cfgget(self._config, _UNIT_SECTION, 'Name')
-        self._runs_on = _parse_runs_on(_cfgget(self._config, _UNIT_SECTION, 'RunsOn'))
+
+    @staticmethod
+    def from_file(fileobj, template_resolver):
+        """Load a ConfigOpts from a configuration file."""
+        _config = load_service_config(fileobj)
```
boegel: why `_config`, this is a local variable?

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+
+        name = _cfgget(_config, _UNIT_SECTION, 'Name')
+        runs_on = _parse_runs_on(_cfgget(_config, _UNIT_SECTION, 'RunsOn'))
+        pre_start_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStartPre', '')
```
boegel: I suspect the last `''` is a default, in case `ExecStartPre` is not defined?

if so, all of these should either i) have a default, ii) have a check to make sure a nice error message is produce if they're not defined but should be

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+        start_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStart')
+        stop_script = _cfgget(_config, _SERVICE_SECTION, 'ExecStop')
+        env = dict([(k, v) for k, v in _config.items(_ENVIRONMENT_SECTION)])
```
boegel: how about just `env = dict(_config.items(_ENVIRONMENT_SECTION))`?

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+
+        return ConfigOpts(name, runs_on, pre_start_script, start_script,
+                stop_script, env, template_resolver)
+
+    def to_params(self, workdir, modules, master_template_args):
+        """Create a ConfigOptsParams object from the ConfigOpts"""
```
boegel: ... from his ConfigOpts instance

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+        return ConfigOptsParams(self.name, self._runs_on, self._pre_start_script,
+                self._start_script, self._stop_script, self._env, workdir,
+                modules, master_template_args)
```
boegel: fix indent, two lines is sufficient?

```python
        return ConfigOptsParams(self.name, self._runs_on, self._pre_start_script, self._start_script,
                                self._stop_script, self._env, workdir, modules, master_template_args)
```

boegel: why not support passing a `ConfigOpts` instance to `ConfigOptsParams` instead, since this looks a bit messy...

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+
+    @staticmethod
+    def from_params(params, template_resolver):
+        """Create a ConfigOpts from a ConfigOptsParams"""
```
boegel: include 'instance' here, twice

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+        return ConfigOpts(params.name, params.runs_on,
+                params.pre_start_script, params.start_script,
+                params.stop_script, params.env, template_resolver)
```
boegel: same remark: two lines + fix indent

hod/config/config.py
```diff
@@ -300,23 +301,56 @@ class ConfigOpts(object):
...
+
+    def __init__(self, name, runs_on, pre_start_script, start_script,
+            stop_script, env, template_resolver):
```
boegel: one line + missing docstring

hod/config/config.py
```diff
@@ -451,3 +498,4 @@ def resolve_config_paths(config, dist):
         raise RuntimeError('A config or a dist must be provided')
 
     return path
+
```
boegel: muh, why?

hod/hodproc.py
```diff
@@ -120,13 +119,21 @@ def distribution(self, *master_template_args, **kwargs):
...
-            self.log.debug('Adding ConfiguredService Task to work with config: %s',
-                    str(config))
-            cfg_opts = ConfigOptsParams(config_filename, m_config.workdir, m_config.modules, master_template_args)
+            self.log.debug('Adding ConfiguredService Task to work with config: %s', str(config))
+            cfg_opts = config.to_params(m_config.workdir, m_config.modules, master_template_args)
             self.tasks.append(Task(ConfiguredService, config.name, ranks_to_run, cfg_opts, master_env))
 
+        if hasattr(self.options, 'script') and self.options.script is not None:
+            script = self.options.script
+            # TODO: How can we test this?
```
boegel: so, does this work as expected?

if you mean "how can we test this in the unit tests", you can mock the `qdel` command to make it do something you can detect (e.g. produce a file in a specific location)

hod/hodproc.py
```diff
@@ -120,13 +119,21 @@ def distribution(self, *master_template_args, **kwargs):
...
+            config = ConfigOpts(script, RUNS_ON_MASTER, '', script + '; qdel $PBS_JOBID', '', master_env, resolver)
```
boegel: move the `+ '; qdel $PBS_JOBID'` bit to where you define `script` (line above this one)

hod/rmscheduler/hodjob.py
```diff
@@ -56,8 +56,7 @@ def __init__(self, options):
 
         self.name_suffix = 'HOD'  # suffixed name, to lookup later
         options_dict = self.options.dict_by_prefix()
-        options_dict['job']['name'] = "%s_%s" % (options_dict['job']['name'],
-                                                self.name_suffix)
+        options_dict['job']['name'] = "%s_%s" % (options_dict['job']['name'], self.name_suffix)
```
boegel: I'd like to see the cluster label here as a default value, makes more sense? if no cluster label is specified, it should be `HOD_foo`, i.e. something short (now it's quite long, I think)?

docs should be updated accordingly, so please sync this branch with `master` now that #43 is merged

hod/subcommands/batch.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+Generate a PBS job script using pbs_python. Will use mympirun to get the cluster
+all started, run a job, and tear down the cluster when it's done.
+
+@author: Stijn De Weirdt (Universiteit Gent)
+@author: Ewan Higgs (Universiteit Gent)
```
boegel: what about me? :( ;)

hod/subcommands/batch.py
```diff
@@ -0,0 +1,103 @@
...
+        }
+        descr = ["Batch options", "Configuration options for the 'batch' subcommand"]
+        self.log.debug("Add config option parser descr %s opts %s", descr, opts)
+        self.add_group_parser(opts, descr)
+
+    def config_options(self):
+        """Add general configuration options."""
+        opts = copy.deepcopy(GENERAL_HOD_OPTIONS)
+        opts.update({
+            'modules': ("Extra modules to load in each service environment", 'string', 'store', None),
```
boegel: `batch --modules` is not covered in the docs yet, please fix

hod/subcommands/batch.py
```diff
@@ -0,0 +1,103 @@
...
+        })
+        descr = ["Create configuration", "Configuration options for the 'batch' subcommand"]
+
+        self.log.debug("Add config option parser descr %s opts %s", descr, opts)
+        self.add_group_parser(opts, descr)
+
+
+class BatchSubCommand(SubCommand):
+    """Implementation of 'create' subcommand."""
```
boegel: batch, not create

hod/subcommands/helptemplate.py
```diff
@@ -44,7 +44,8 @@ class HelpTemplateOptions(GeneralOption):
 
 def mk_registry():
     """Make a TemplateRegistry and register basic items"""
-    config_opts = ConfigOptsParams(filename='hod.conf', workdir='WORKDIR',
+    config_opts = ConfigOptsParams('svc-name', 'MASTER', 'ExecPreStart',
+            'ExecStart', 'ExecStop', dict(), workdir='WORKDIR',
         modules=['MODULES'], master_template_kwargs=[])
```
boegel: two lines + fix indent

hod/config/config.py
ehiggs: Just adding a space on either side of `=` to placate pep checkers.

hod/config/config.py
ehiggs: Well if they need to be defined, then they will be defined. Not sure what you mean by `ii` there.

I think having defaults here will be messy since `template_resolver` can't really have a default. So pushing it forward in the list of parameters would mix it in with the other params that come from the actual config. The cleanest, imo, is to make another `staticmethod` `with_defaults` which only takes the `name, runs_on, start_script, env, and template_resolver`.

hod/config/config.py
ehiggs: :+1:

hod/config/config.py
ehiggs: `ConfigOptsParams` only exists to be a picklable version of `ConfigOpts`. Probably a better solution would be to register a pickling function and drop the whole notion of moving between the types. That's not in scope for this PR, imo.

hod/config/config.py
ehiggs: I can make it one line, but docstring on init is unneeded imo. There is a class level docstring.

hod/hodproc.py
ehiggs: We can't really mock out qdel as a command. We can mock out the system of using subprocess.

hod/hodproc.py
ehiggs: :+1:

hod/config/config.py
boegel: Sure, but break it up while you're touching it (I feel I'm being reasonable w.r.t. style remarks, so bare with me).

hod/config/config.py
```diff
@@ -300,23 +301,52 @@ class ConfigOpts(object):
...
-        self._config = load_service_config(fileobj)
-        self.name = _cfgget(self._config, _UNIT_SECTION, 'Name')
-        self._runs_on = _parse_runs_on(_cfgget(self._config, _UNIT_SECTION, 'RunsOn'))
+
+    @staticmethod
+    def from_file(fileobj, template_resolver):
+        """Load a ConfigOpts from a configuration file."""
+        config = load_service_config(fileobj)
+
+        name = _cfgget(config, _UNIT_SECTION, 'Name')
```
boegel: what would the `_cfgget` call result in if `Name` is not available? a proper error message, a traceback?

if the latter, I'd prefer something like:

```python
name = _cfgget(config, _UNIT_SECTION, 'Name', None)
if name is None:
    raise ...
```

hod/rmscheduler/hodjob.py
ehiggs: Sure. I can drop `--job-name`. Otherwise I think it gets confusing which name should be used in the job.

1. If a label is given; job name is `HOD_<label>`
2. If a label is not given, job name is `HOD_job`

What do you think?

boegel: Sure, works for me! :+1:

hod/config/config.py
ehiggs: It raises `NoOptionError`

ehiggs: I added a test to demonstrate this raises `NoOptionError`. Check 9c79e83.

hod/subcommands/batch.py
ehiggs: Addded.

hod/subcommands/batch.py
```diff
@@ -0,0 +1,103 @@
...
+            sys.stderr.write('Missing config options. Exiting.\n')
+            return 1
+
+        if optparser.options.script is None:
+            sys.stderr.write('Missing script. Exiting.\n')
+            return 1
+
+        try:
+            j = PbsHodJob(optparser)
+            j.run()
```
boegel: I'm seeing this, even though the cluster is submitted:

```
Jobs submitted: []
```

ehiggs: This was a bug with the job filter lookup. `HOD` went from being a suffix to a prefix so a filter of `HOD$` was wrong.

hod/rmscheduler/hodjob.py
```diff
@@ -54,14 +54,18 @@ def __init__(self, options):
...
         self.set_type_class()
 
-        self.name_suffix = 'HOD'  # suffixed name, to lookup later
+        # HOD_<label> if label is given; HOD_job otherwise.
+        self.name_prefix = 'HOD'
         options_dict = self.options.dict_by_prefix()
-        options_dict['job']['name'] = "%s_%s" % (options_dict['job']['name'],
-                                                self.name_suffix)
+        label = 'job'
+        if 'label' in options_dict[''] and not options_dict['']['label'] is None:
```
boegel: just use `options.label` here?

boegel: https://github.com/ehiggs/hanythingondemand/pull/18

PR #68, opened 2015-09-03
add guard around mpi4py import in hod.local

boegel:
without this, `create --help` doesn't work unless `mpi4py` is available (since it imports from `hod.local`), which is kinda stupid

### DIFF
--- a/hod/local.py
+++ b/hod/local.py
@@ -33,7 +33,6 @@ import os
 import random
 import string
 import sys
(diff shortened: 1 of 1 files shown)


ehiggs:
`hod batch` needs this too.

boegel:
OK, please add it there too then.

Maybe we can hoist this into a function, would that work?

PR #70, opened 2015-09-10
add support for --hod-module + some refactoring + docs update

boegel:
@ehiggs: This enables me to submit to golett from the login nodes, using the system Python, as follows:

```
hod create --workdir=$VSC_SCRATCH/hod --dist=IPython-notebook-3.2.1 --hod-module=hanythingondemand/3.0.0dev-intel-2015a-Python-2.7.10
```

Seems to be working (I can connect to the IPython web UI for a cluster that was submitted like this).

### DIFF
--- a/docs/Command_line_interface.rst
+++ b/docs/Command_line_interface.rst
@@ -108,11 +108,10 @@ Configuration options for ``hod create``
 ``hod create --hod-module <module name>``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-*required*
+**must be specified**
 
-Specify the ``hanythingondemand`` module that must be loaded in the job that is submitted for the HOD cluster.
-
-Can also be specified via ``$HOD_CREATE_HOD_MODULE``.
+Specify the ``hanythingondemand`` module that must be loaded in the job that is submitted for the HOD cluster;
+can also be specified via ``$HOD_CREATE_HOD_MODULE``.
 
 
 .. _cmdline_create_options_workdir:
@@ -120,11 +119,9 @@ Can also be specified via ``$HOD_CREATE_HOD_MODULE``.
 ``hod create --workdir <path>``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-*required*
+**must be specified**
 
-Specify the top-level working directory to use.
-
-Can also be specified via ``$HOD_CREATE_WORKDIR``.
+Specify the top-level working directory to use; can also be specified via ``$HOD_CREATE_WORKDIR``.
 
 
 .. _cmdline_create_options_hodconf:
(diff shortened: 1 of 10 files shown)


hod/rmscheduler/hodjob.py
```diff
@@ -118,28 +118,13 @@ class PbsHodJob(MympirunHod):
...
-                        ebmodname_envvar)
-
-        # FIXME
-        self.modules.append(ebmodname)
+
+        if options.options.hod_module:
+            self.modules = [options.options.hod_module]
+        else:
+            sys.stderr.write("HOD module to load in job is not specified (--hod-module).\n")
+            sys.exit(1)
```
ehiggs: Can we not throw here?

hod/node/node.py
```diff
@@ -41,17 +39,49 @@
 
 from hod.commands.command import ULimit
 
+try:
```
ehiggs: This again!?

docs/Command_line_interface.rst
```diff
@@ -117,22 +95,67 @@ Create a hanythingondemand cluster, with the specified label (optional) and clus
...
 
+.. _cmdline_create_options_hod_module:
+
+``hod create --hod-module <module name>``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**must be specified**
+
+Specify the ``hanythingondemand`` module that must be loaded in the job that is submitted for the HOD cluster;
+can also be specified via ``$HOD_CREATE_HOD_MODULE``.
```
ehiggs: I want to see an associated EasyBuild file that sets this so it's always in the environment when someone's using it.

ehiggs:
https://github.com/boegel/hanythingondemand/pull/6 addresses my comments and fixes tests.

hod/node/node.py
boegel: if we want `hod create` and `hod batch` to work with only the system Python on our systems, we need this

unless we install `netifaces` and `netaddr` just to make the `import's work, which is stupid imho since `netifaces` and `netaddr` are only needed in the HOD 'backend', not for interactive use of `hod`

docs/Command_line_interface.rst
boegel: I'm planning exactly this: the module we'll put in place for interactive use on the login node should define `$HOD_CREATE_HOD_MODULE` and `$HOD_CREATE_WORKDIR` (+ the same for the `batch` subcommand)

this will make it a lot easier for users

hod/rmscheduler/hodjob.py
boegel: if it's OK to require `--hod-module` for `genconfig`, yes, this can move to `validate_pbs_option`

hod/node/node.py
ehiggs: Sure. Then let's lift all these decorator checks into a common module checker, if possible.

PR #74, opened 2015-09-11
include label in localworkdir, if available

boegel:


boegel:
Jenkins: test this please

ehiggs:
I'm not sure what the benefit is here. If it's to help `hod clean` determine the localworkdir, that's better done by making another `hod.d/<label>/` file.

boegel:
It's mostly so people can easily figure out which local workdir corresponds to which cluster, if they ever need to (e.g. when going to the HOD workdir, and navigating via `ls` and co).

It's not intended to make internal functionality easier.

boegel:
Jenkins: test this please

PR #84, opened 2015-09-15
Fix for genconfig for issue #83

ehiggs:
fix for #83.


boegel:
no change in the unit tests to make sure we catch this in the future?

ehiggs:
Unit test added. :shipit:

boegel:
:+1:

PR #87, opened 2015-09-21
Batch fixes

ehiggs:
Send batch job output to files.

Give batch jobs unlimited timeouts. Using a simple wordcount job didn't catch the fact that batch jobs had a 120 second timeout as the rest of the services (which are sent to the bg).

### DIFF
--- a/hod/commands/command.py
+++ b/hod/commands/command.py
@@ -43,6 +43,7 @@ from vsc.utils import fancylogger
 
 
 COMMAND_TIMEOUT = 120  # timeout
+NO_TIMEOUT = -1
 
 
 class Command(object):
@@ -122,15 +123,20 @@ class Command(object):
         timedout = False
         while p.poll() is None:
             if os.path.exists("/proc/%s" % (p.pid)):
-                now = datetime.datetime.now()
-                if (now - start).seconds > self.timeout:
-                    if timedout is False:
-                        os.kill(p.pid, signal.SIGTERM)
-                        self.log.debug("Timeout occured with cmd %s. took more than %i secs to complete.",
-                                self.command, self.timeout)
-                        timedout = True
-                    else:
-                        os.kill(p.pid, signal.SIGKILL)
+                if self.timeout == NO_TIMEOUT:
+                    self.log.debug("There is no timeout for cmd %s and it is still running" % self.command)
+                    time.sleep(10)
+                    continue
+                else:
+                    now = datetime.datetime.now()
+                    if (now - start).seconds > self.timeout:
+                        if timedout is False:
+                            os.kill(p.pid, signal.SIGTERM)
+                            self.log.debug("Timeout occured with cmd %s. took more than %i secs to complete.",
+                                    self.command, self.timeout)
+                            timedout = True
+                        else:
+                            os.kill(p.pid, signal.SIGKILL)
             time.sleep(1)
 
         if self.fake_pty:
--- a/hod/commands/command.py
+++ b/hod/commands/command.py
@@ -42,8 +42,11 @@ import time
 from vsc.utils import fancylogger
 
 
-COMMAND_TIMEOUT = 120  # timeout
-NO_TIMEOUT = -1
+COMMAND_TIMEOUT = 120  # timeout in seconds
+NO_TIMEOUT = None # No timeout
+
+TIMEOUT_POLL_TIME = 1
+NO_TIMEOUT_POLL_TIME = 10
 
 
 class Command(object):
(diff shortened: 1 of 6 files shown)


ehiggs:
This is based on the relabel branch.

hod/commands/command.py
```diff
@@ -43,6 +43,7 @@
 
 
 COMMAND_TIMEOUT = 120  # timeout
+NO_TIMEOUT = -1
```
boegel: use `None`?

hod/commands/command.py
```diff
@@ -122,15 +123,20 @@ def run(self):
         timedout = False
         while p.poll() is None:
```
boegel: why are we reinventing what's available in `vsc.utils.run` here?

hod/commands/command.py
```diff
@@ -122,15 +123,20 @@ def run(self):
...
-                    if timedout is False:
-                        os.kill(p.pid, signal.SIGTERM)
-                        self.log.debug("Timeout occured with cmd %s. took more than %i secs to complete.",
-                                self.command, self.timeout)
-                        timedout = True
-                    else:
-                        os.kill(p.pid, signal.SIGKILL)
+                if self.timeout == NO_TIMEOUT:
+                    self.log.debug("There is no timeout for cmd %s and it is still running" % self.command)
+                    time.sleep(10)
```
boegel: magic value, make this a constant?

repolling seems a bit frequent, especially because we're not doing anything at all here (like, fetching output)?

hod/hodproc.py
```diff
@@ -94,6 +95,17 @@ def _setup_template_resolver(m_config, master_template_args):
...
 
+def _script_output_paths(script_name):
+    """
+    Given a script path, return the path to the output files. This uses a
+    sceheme mirroring how PBS actually works where jobs have
+    '$PBS_O_WORKDIR/<script-name>.[eo]<$PBS_JOBID>'
+    """
+    script_basename = os.path.basename(script_name)
+    script_stdout = mkpath('$PBS_O_WORKDIR', '%s.o${PBS_JOBID}' % script_basename)
+    script_stderr = mkpath('$PBS_O_WORKDIR', '%s.e${PBS_JOBID}' % script_basename)
```
boegel: I would also include the cluster label?

`hod-${label}-${script_basename}.o$PBS_JOBID` ?

hod/hodproc.py
```diff
@@ -129,9 +141,13 @@ def distribution(self, *master_template_args, **kwargs):
             self.tasks.append(Task(ConfiguredService, config.name, ranks_to_run, cfg_opts, master_env))
 
         if hasattr(self.options, 'script') and self.options.script is not None:
-            script = self.options.script + '; qdel $PBS_JOBID'
+            script = self.options.script
+            # Use scring concatenation since os.path.join will expand envvars, which we don't want!
```
boegel: scring?

boegel: wow, dafuq, you're right, os.path.join *does* expand env vars... O_o

```
$ python -c "import os; print os.path.join('$HOME', 'foo')"
/Users/kehoste/foo
```

hod/commands/command.py
boegel: if we use one of the `run*` functions available in `vsc.utils.run` we can probably trash dozens of lines here...

boegel: @ehiggs: open an issue for this, it shouldn't block this PR

ehiggs: #90 opened.

hod/hodproc.py
ehiggs: No, that's your shell expanding it.

I was an idiot when I ran into that issue and needed to `basename` the script. That's why I was getting a path attached to the script.

hod/hodproc.py
```diff
@@ -94,6 +95,22 @@ def _setup_template_resolver(m_config, master_template_args):
...
+    Given a script path, return the path to the output files. This uses a
+    sceheme mirroring how PBS actually works where jobs have
+    '$PBS_O_WORKDIR/<script-name>.[eo]<$PBS_JOBID>'
+    """
+    script_basename = os.path.basename(script_name)
+    if label is None:
+        script_stdout = mkpath('$PBS_O_WORKDIR', 'hod-%s.o${PBS_JOBID}' % script_basename)
+        script_stderr = mkpath('$PBS_O_WORKDIR', 'hod-%s.e${PBS_JOBID}' % script_basename)
+    else:
+        script_stdout = mkpath('$PBS_O_WORKDIR', 'hod-%s-%s.o${PBS_JOBID}' % (label, script_basename))
```
boegel: there's a lot of copy-pasting here...

how about:

```python
if label is None:
    output_label = 'hod-%s' % os.path.basename(script_name)
else:
    output_label = 'hod-%s-%s' % (label, os.path.basename(script_name))

script_stdout = mkpath('$PBS_O_WORKDIR', '%s.o${PBS_JOBID}' % output_label)
script_stderr = mkpath('$PBS_O_WORKDIR', '%s.e${PBS_JOBID}' % output_label)
return (script_stdout, script_stderr)
```

also, there's no unit test added for this function?

PR #89, opened 2015-09-24
Feature/h base

### LINKED ISSUE #81: support HBase?
cc @molden

molden:
fixes #81

### DIFF
--- a/etc/hod/HBase-1.0.2/datanode.conf
+++ b/etc/hod/HBase-1.0.2/datanode.conf
@@ -1,0 +1,14 @@
+#-*- cfg -*-
+# vim: ft=cfg
+[Unit]
+Name=datanode
+RunsOn=all
+
+[Service]
+ExecStart=$$EBROOTHADOOP/sbin/hadoop-daemon.sh start datanode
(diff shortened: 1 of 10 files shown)


boegel:
@molden: great, thanks! I'll let @ehiggs take a look at this (and I'll tackle https://github.com/hpcugent/easybuild-easyconfigs/pull/1990)

molden:
@ehiggs : as you may notice for HBase it is also necessary to launch the HDFS filesystem to work as HBase depends on it or at least it depends on a filesystem that implements the HDFS API. So in theory it should also work with a version of GPFS that implements the right APIs, if I'm not mistaken?

molden:
@ehiggs : I could reproduce the error: 'port already in use' for the regionserver. So it is related to this bug in HBase: https://issues.apache.org/jira/browse/HBASE-13479. A workaround is submitted to the PR. I have tested this and it works.

ehiggs:
Thanks for the fix to the patch. I've tested it by running the tutorial examples [here](http://akbarahmed.com/2012/08/13/hbase-command-line-tutorial/).

Thanks for the patch!

PR #92, opened 2015-09-30
Use PBS_JOBID in the localworkdir to group localworkdirs together.

ehiggs:
This lets us search logs more easily across nodes, and remove old cluster files
since we know they're all under the jobid dir.

### DIFF
--- a/hod/config/config.py
+++ b/hod/config/config.py
@@ -37,7 +37,7 @@ from pkg_resources import Requirement, resource_filename, resource_listdir
 import hod
 from hod.node.node import Node
 from hod.commands.command import COMMAND_TIMEOUT
-from hod.config.template import mklocalworkdir
+import hod.config.template as hct
 
 
 from vsc.utils import fancylogger
@@ -187,7 +187,7 @@ class PreServiceConfigOpts(object):
 
     @property
     def localworkdir(self):
-        return mklocalworkdir(self.workdir)
+        return hct.mklocalworkdir(self.workdir)
 
     @property
     def configdir(self):
@@ -364,7 +364,7 @@ class ConfigOpts(object):
 
     @property
     def localworkdir(self):
-        return mklocalworkdir(self._tr.workdir)
+        return hct.mklocalworkdir(self._tr.workdir)
 
     @property
     def configdir(self):
--- a/hod/config/config.py
+++ b/hod/config/config.py
@@ -363,8 +363,8 @@ class ConfigOpts(object):
         return self._tr.workdir
 
     @property
-    def localworkdir(self, label=None):
-        return mklocalworkdir(self._tr.workdir, label=None)
+    def localworkdir(self):
+        return mklocalworkdir(self._tr.workdir)
 
     @property
     def configdir(self):
--- a/hod/config/template.py
+++ b/hod/config/template.py
@@ -105,18 +105,17 @@ def register_templates(template_registry, config_opts):
     for ct in templates:
         template_registry.register(ct)
 
-def mklocalworkdir(workdir, label=None):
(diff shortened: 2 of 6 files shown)


hod/config/template.py
```diff
@@ -105,18 +105,17 @@ def register_templates(template_registry, config_opts):
...
 
-def mklocalworkdir(workdir, label=None):
+def mklocalworkdir(workdir):
     '''
     Construct the pathname for a workdir with a path local to this
     host/job/user.
     '''
     user = _current_user()
     pid = os.getpid()
+    jobid = os.getenv('PBS_JOBID', '')
```
boegel: use something liek `PBS_JOBID_undefined` rather than empty string as fallback value?

boegel: or, hmm, not sure, but just ignoring `PBS_JOBID` not being defined is maybe not the best thing to do?

hod/config/config.py
```diff
@@ -363,8 +363,8 @@ def workdir(self):
         return self._tr.workdir
 
     @property
-    def localworkdir(self, label=None):
-        return mklocalworkdir(self._tr.workdir, label=None)
+    def localworkdir(self):
```
boegel: changing the signature here also means we need to adjust the location where it was used with `label=`?

hod/config/template.py
```diff
@@ -105,18 +105,17 @@ def register_templates(template_registry, config_opts):
...
     hostname = socket.getfqdn()
     dir_name = '.'.join([user, hostname, str(pid)])
-    if label:
-        dir_name = '%s.%s' % (label, dir_name)
```
boegel: leave this here, just in case we reinstate using the labelled localworkdir?

ehiggs: It's pretty easy to put back so I don't think it's needed.

hod/config/config.py
ehiggs: I'm pretty sure it wasn't used. Can you find where it's allegedly used?

hod/config/template.py
ehiggs: `PBS_JOBID_undefined` won't be unique and no other value really makes sense, imo. I could remove the default and force it to have a `PBS_JOBID`.

hod/config/template.py
boegel: OK, fine. :-)

hod/config/config.py
boegel: I guess that explains why the test passes, but it doesn't actually work ^_^

hod/config/template.py
boegel: yes please, check whether `jobid` is `None`, and raise an error if it is to complain that `$PBS_JOBID` is not defined

ehiggs: >raise an error if it is to complain that $PBS_JOBID is not defined

I'll just let getenv raise and pass through.

boegel: getenv doesn't raise anything it returns `None`, and you'll get an error when using `jobid` below for using `None`

so, raise a clear error instead, if `jobid` is `None`

hod/config/template.py
```diff
@@ -105,18 +105,19 @@ def register_templates(template_registry, config_opts):
...
+def mklocalworkdir(workdir):
     '''
     Construct the pathname for a workdir with a path local to this
     host/job/user.
     '''
     user = _current_user()
     pid = os.getpid()
+    jobid = os.getenv('PBS_JOBID')
+    if jobid is None:
+        raise RuntimeError('PBS_JOBID must be defined to create a localworkdir.')
```
boegel: make this `$PBS_JOBID`, to make it very clear we're talking about an environment variable here...

PR #85, opened 2015-09-15
Document ipython logging; use HOD_LOCALWORKDIR in docs. doc hod clean.

ehiggs:
Implements #79, #76, and adds ipython logging documentation.

PR #95, opened 2015-10-07
Hod clean removes localworkdir

ehiggs:
With this PR, hod clean will remove localworkdirs.

### DIFF
--- a/hod/cluster.py
+++ b/hod/cluster.py
@@ -28,6 +28,7 @@ Cluster and label functions.
 @author: Kenneth Hoste (Universiteit Gent)
 """
 
+import errno
 import os
 import sys
 import shutil
@@ -279,6 +280,7 @@ def clean_cluster_info(master, cluster_info):
     """
     for info in cluster_info:
         if info.pbsjob is None and info.jobid.endswith(master):
+            rm_cluster_localworkdir(info.label)
             rm_cluster_info(info.label)
 
 
@@ -297,12 +299,19 @@ def rm_cluster_info(label):
 
 def rm_cluster_localworkdir(label):
     """Remove a cluster localworkdir directory"""
-    info_dir = os.path.join(cluster_info_dir(), label)
     jobid = cluster_jobid(label)
     workdir = cluster_workdir(label)
-    jobid_workdir = os.path.join(workdir, jobid)
-    shutil.rmtree(jobid_workdir)
-    print 'Removed cluster localworkdir directory %s for cluster labeled %s' % (jobid_workdir, label)
+    jobid_workdir = os.path.join(workdir, 'hod', jobid)
+    try:
+        shutil.rmtree(jobid_workdir)
+        print 'Removed cluster localworkdir directory %s for cluster labeled %s' % (jobid_workdir, label)
+    except OSError, e:
+        # If we cancelled a queued job we might not have a localworkdir yet.
+        if e.errno == errno.ENOENT:
+            print 'Note: No localworkdir directory %s for cluster labeled %s' % (jobid_workdir, label)
+            print 'Maybe this was a cancelled queued job'
+        else:
+            raise
 
 
 def mv_cluster_info(label, newlabel):
--- a/hod/cluster.py
+++ b/hod/cluster.py
@@ -256,7 +256,7 @@ def mk_cluster_info(label, jobid, workdir):
 def save_cluster_info(cluster_info):
     """
     Save info (job ID, env script, ...) for this cluster in the cluster info dir.
-    cluster_info is a dict created by `gen_cluster_info` (not a ClusterInfo)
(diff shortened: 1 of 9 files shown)


boegel:
@ehiggs: please sync with master once #85 is merged, and update the section on `hod clean` to mention that it also cleans up the local work dirs

ehiggs:
Jenkins: test this please.

boegel:
@ehiggs: please sync with master to clean up a diff?

hod/cluster.py
```diff
@@ -149,6 +149,11 @@ def cluster_jobid(label):
     return open(_cluster_info(label, 'jobid')).read()
 
 
+def cluster_workdir(label):
+    """Return workdir for cluster with specified label."""
+    return open(_cluster_info(label, 'workdir')).read()
```
boegel: what if what you're trying to read doesn't exist?

be a little bit more careful?

```python
txt = None
filepath = _cluster_info(label, 'workdir')
if os.path.exists(filepath):
    try:
        txt = open(filepath).read()
    except IOError as err:
        raise SomeError("Unexpected error occured while trying to read %s: %s" % (filepath, err))
else:
    raise SomeError("some msg")

return txt
```

hod/cluster.py
```diff
@@ -227,18 +232,26 @@ def mk_cluster_info(label, jobid):
...
             jobid_file.write(jobid)
     except IOError as err:
-        _log.error("Failed to write cluster info files: %s", err)
+        _log.error("Failed to write jobid file: %s", err)
+
+    try:
+        with open(os.path.join(info_dir, 'workdir'), 'w') as workdir_file:
+            workdir_file.write(workdir)
+    except IOError as err:
+        _log.error("Failed to write workdir file: %s", err)
```
boegel: `log.error` doesn't exit or raise here, does it?

hod/cluster.py
boegel: oh, and use `with open`

hod/cluster.py
```diff
@@ -227,18 +232,26 @@ def mk_cluster_info(label, jobid):
...
 
 
 def save_cluster_info(cluster_info):
-    """Save info (job ID, env script, ...) for this cluster in the cluster info dir."""
+    """
+    Save info (job ID, env script, ...) for this cluster in the cluster info dir.
+    cluster_info is a dict created by `gen_cluster_info` (not a ClusterInfo)
```
boegel: `not a ClusterInfo instance` ?

hod/cluster.py
```diff
@@ -271,6 +285,22 @@ def rm_cluster_info(label):
     print 'Removed cluster info directory %s for cluster labeled %s' % (info_dir, label)
 
 
+def rm_cluster_localworkdir(label):
+    """Remove a cluster localworkdir directory"""
+    jobid = cluster_jobid(label)
+    workdir = cluster_workdir(label)
+    jobid_workdir = os.path.join(workdir, 'hod', jobid)
+    try:
+        shutil.rmtree(jobid_workdir)
```
boegel: only try to remove it if it exists? `os.path.exists(jobid_workdir)` ?

hod/cluster.py
```diff
@@ -28,6 +28,7 @@
 @author: Kenneth Hoste (Universiteit Gent)
 """
 
+import errno
```
boegel: not used, please drop

hod/cluster.py
boegel: discussed with @ehiggs off-pr, OK as is

This should never be used without `workdir` being there; if it is, it's a bug that will manifest itself as an IOError being raised, so fine.

hod/cluster.py
boegel: ping?

hod/cluster.py
```diff
@@ -289,6 +316,8 @@ def post_job_submission(label, jobs):
     elif len(jobs) > 1:
         sys.stderr.write('Warning: More than one job found: %s\n' % str([j.jobid for j in jobs]))
     job = jobs[0]
-    print "Jobs submitted: %s" % str(job)
-    mk_cluster_info(label, job.jobid)
-
+    print "Job submitted: %s" % str(job)
+    try:
+        mk_cluster_info(label, job.jobid, workdir)
+    except OSError:
```
boegel: IOError, mention the error, and exit > 0

```python
except (IOError, OSError) as err:
    sys.stderr.write("...: %s\n" % (label, err))
    sys.exit(1)
```

hod/local.py
```diff
@@ -87,7 +87,11 @@ def main(args):
 
         _log.debug("Creating cluster info using label '%s'", label)
         cluster_info = gen_cluster_info(label, optparser.options)
-        save_cluster_info(cluster_info)
+        try:
+            save_cluster_info(cluster_info)
+        except OSError, e:
```
boegel: here too, (also) catch IOError

PR #99, opened 2015-10-13
Refactor list and helptemplate printing to hod.table.

### LINKED ISSUE #98: hod list table text
Currently `hod list` prints label, jobid and then a job state. But the job state prints the job id again, followed by job state (Q - queued, R - running, etc) and then the hosts. The job state and hosts should be broken out into their own columns.

ehiggs:
This should fix #98.

### DIFF
--- a/hod/subcommands/helptemplate.py
+++ b/hod/subcommands/helptemplate.py
@@ -28,9 +28,11 @@ Print out template parameters used by hod.
 @author: Ewan Higgs (Ghent University)
 @author: Kenneth Hoste (Ghent University)
 """
+import os
 from vsc.utils.generaloption import GeneralOption
 
 import hod.config.template as hct
+import hod.table as ht
 from hod import VERSION as HOD_VERSION
 from hod.subcommands.subcommand import SubCommand
 from hod.mpiservice import master_template_opts
@@ -56,15 +58,14 @@ def mk_registry():
     return reg
 
 
-def mk_fmt_str(fields, resolver):
-    """Calculate the format string for printing the template parameters."""
-    longest_name = max(fields.values(), key=lambda x: len(x.name)).name
-    max_name_len = len(longest_name)
-    value_names = map(resolver, ['$%s'% v.name for v in fields.values()])
-    longest_value = max(value_names, key=len)
-    max_val_len = len(longest_value)
-    return '%%-%ds:\t%%-%ds\t%%s' % (max_name_len, max_val_len)
-
+def format_rows(fields, resolver):
+    """
+    Given a list of fields, return the template substituted versions of the text
+    """
+    rows = []
+    for v in sorted(fields.values(), key=lambda x: x.name):
+        rows.append((v.name, resolver('$%s' % v.name), v.doc))
+    return rows
 
 class HelpTemplateSubCommand(SubCommand):
     """Implementation of 'help-template' subcommand."""
@@ -74,11 +75,11 @@ class HelpTemplateSubCommand(SubCommand):
 
     def run(self, args):
         """Run 'help-template' subcommand."""
+        os.environ['PBS_JOBID'] = '123.%s' % os.environ['PBS_DEFAULT']
         optparser = HelpTemplateOptions(go_args=args, envvar_prefix=self.envvar_prefix, usage=self.usage_txt)
         reg = mk_registry()
         resolver = hct.TemplateResolver(**reg.to_kwargs())
         print 'Hanythingondemand template parameters'
-        fmt_str = mk_fmt_str(reg.fields, resolver)
-        print fmt_str % ('Parameter name', 'Value', 'Documentation')
-        for v in sorted(reg.fields.values(), key=lambda x: x.name):
-            print fmt_str % (v.name, resolver('$%s' % v.name), v.doc)
+        headers = ('Parameter name', 'Value', 'Documentation')
+        formatted_rows = format_rows(reg.fields, resolver)
+        print ht.format_table(formatted_rows, headers)
--- a/hod/subcommands/helptemplate.py
+++ b/hod/subcommands/helptemplate.py
@@ -38,6 +38,7 @@ from hod.subcommands.subcommand import SubCommand
 from hod.mpiservice import master_template_opts
 from hod.config.config import ConfigOptsParams
 from hod.commands.command import COMMAND_TIMEOUT
+from hod.utils import setup_diagnostic_environment
 
 
 class HelpTemplateOptions(GeneralOption):
@@ -75,10 +76,7 @@ class HelpTemplateSubCommand(SubCommand):
 
     def run(self, args):
         """Run 'help-template' subcommand."""
-        # We need PBS_JOBID defined in the environment to generate localworkdirs
-        if 'PBS_DEFAULT' not in os.environ:
-            os.environ['PBS_DEFAULT'] = 'pbs-master'
-        os.environ['PBS_JOBID'] = '123.%s' % os.environ['PBS_DEFAULT']
+        setup_diagnostic_environment()
 
         _ = HelpTemplateOptions(go_args=args, envvar_prefix=self.envvar_prefix, usage=self.usage_txt)
         reg = mk_registry()
--- a/hod/subcommands/helptemplate.py
+++ b/hod/subcommands/helptemplate.py
@@ -63,8 +63,8 @@ def format_rows(fields, resolver):
     Given a list of fields, return the template substituted versions of the text
     """
     rows = []
-    for v in sorted(fields.values(), key=lambda x: x.name):
-        rows.append((v.name, resolver('$%s' % v.name), v.doc))
+    for val in sorted(fields.values(), key=lambda x: x.name):
+        rows.append((val.name, resolver('$%s' % val.name), val.doc))
     return rows
 
 class HelpTemplateSubCommand(SubCommand):
@@ -75,8 +75,12 @@ class HelpTemplateSubCommand(SubCommand):
 
     def run(self, args):
         """Run 'help-template' subcommand."""
+        # We need PBS_JOBID defined in the environment to generate localworkdirs
+        if 'PBS_DEFAULT' not in os.environ:
+            os.environ['PBS_DEFAULT'] = 'pbs-master'
         os.environ['PBS_JOBID'] = '123.%s' % os.environ['PBS_DEFAULT']
-        optparser = HelpTemplateOptions(go_args=args, envvar_prefix=self.envvar_prefix, usage=self.usage_txt)
+
+        _ = HelpTemplateOptions(go_args=args, envvar_prefix=self.envvar_prefix, usage=self.usage_txt)
         reg = mk_registry()
         resolver = hct.TemplateResolver(**reg.to_kwargs())
         print 'Hanythingondemand template parameters'
--- a/hod/subcommands/listcmd.py
+++ b/hod/subcommands/listcmd.py
@@ -49,21 +49,17 @@ class ListOptions(GeneralOption):
     """Option parser for 'list' subcommand."""
     VERSION = HOD_VERSION
 
-def mk_fmt_str(labels, jobid, header):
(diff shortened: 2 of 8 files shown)


hod/subcommands/helptemplate.py
```diff
@@ -74,11 +75,11 @@ class HelpTemplateSubCommand(SubCommand):
 
     def run(self, args):
         """Run 'help-template' subcommand."""
+        os.environ['PBS_JOBID'] = '123.%s' % os.environ['PBS_DEFAULT']
```
boegel: say what? :)

hod/subcommands/helptemplate.py
```diff
@@ -56,15 +58,14 @@ def mk_registry():
...
-    longest_value = max(value_names, key=len)
-    max_val_len = len(longest_value)
-    return '%%-%ds:\t%%-%ds\t%%s' % (max_name_len, max_val_len)
-
+def format_rows(fields, resolver):
+    """
+    Given a list of fields, return the template substituted versions of the text
+    """
+    rows = []
+    for v in sorted(fields.values(), key=lambda x: x.name):
```
boegel: no single-letter variables outside of list comprehensions please (the one in the lambda is OK too though)

hod/subcommands/listcmd.py
```diff
@@ -55,14 +56,15 @@ def mk_fmt_str(labels, jobid, header):
     return '%%-%ds\t%%-%ds\t%%s' % (max_label_len, max_jobid_len)
```
boegel: kill this function, no longer used?

hod/subcommands/listcmd.py
```diff
@@ -55,14 +56,15 @@ def mk_fmt_str(labels, jobid, header):
...
 
 
-def format_cluster_info(cluster_info):
-    """Turn a list of 'ClusterInfo' objects into a 2-tuple of strings."""
+def format_rows(cluster_info):
```
boegel: don't use the same name as the function in `helptemplate.py`, it will get very confusing, so rename them both (`format_list_rows` here?)

hod/subcommands/listcmd.py
```diff
@@ -55,14 +56,15 @@ def mk_fmt_str(labels, jobid, header):
...
+    """Turn a list of 'ClusterInfo' objects into a list of strings."""
     ret = []
     for info in cluster_info:
         label = info.label if info.label is not None else '<no-label>'
         jobid = info.jobid
-        job = str(info.pbsjob) if info.pbsjob is not None else '<job-not-found>'
-        ret.append((label, jobid, job))
+        state = info.pbsjob.state if info.pbsjob is not None else '<job-not-found>'
+        hosts = info.pbsjob.hosts if info.pbsjob is not None else '<no-hosts-found>'
```
boegel: maybe this is clearer?

```python
if info.pbsjob is not None:
    state, hosts = info.pbsjob.state, info.pbsjob.hosts
else:
    state, hosts = '<job-not-found>', '(none)'
```

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+        cols.append([row[col] for row in rows])
+    return cols
+
+def format_table(rows, headers=None):
+    """
+    Formats a table of data. 
+    Params
+    ------
+    rows - list of lists containing each row of data
+    headers - None or list of strings for the table header.
```
boegel: use `@param rows`, @param headers`?

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+
+    The number of headers must be the same as the number of columns or None.
+    """
+
+    if not rows:
+        raise ValueError('No rows to print')
+
+    if headers is not None and len(headers) != len(rows[0]):
```
boegel: maybe also check whether all rows have same length?

`nub(map(len, rows)) == len(rows[0])`

`nub` is available in `vsc.utils.missing`

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+        out = fmt_str % tuple(headers)
+        out += '\n'
+    else:
+        cols = _pivot(rows)
+        fmt_str = _mk_fmt_str(cols)
+
+    for i, line in enumerate(rows):
+        out += fmt_str % tuple(line)
+        if i != len(rows)-1:
+            out += '\n'
```
boegel: the `i != len` bit is a bit much, no?

```python
out += ''.join([fmt_str % tuple(line) for line in rows])
out += '\n'
```

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if headers is not None and len(headers) != len(rows[0]):
+        raise ValueError('Headers and rows must have the same length')
+
+    out = ''
+    if headers:
+        cols = _pivot([headers] + rows)
+        fmt_str = _mk_fmt_str(cols)
+        out = fmt_str % tuple(headers)
+        out += '\n'
```
boegel: collapse with previous line, and use `+=` on line above

```python
out += fmt_str % tuple(headers) + '\n'
```

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+    """Given a list of rows, pivot it into a list of columns - or vice versa."""
+    if not rows:
+        return []
+
+    cols = []
+    ncols = len(rows[0])
+    for col in range(ncols):
+        cols.append([row[col] for row in rows])
+    return cols
+
```
boegel: two empty lines between top-level functions

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+def _mk_fmt_str(fields):
+    """Calculate the format string for printing the template parameters."""
+    field_lens = []
+    for field in fields:
+        field_lens.append(max(map(len, field)))
+    fmt_str = ""
+    for i, field_len in enumerate(field_lens):
+        fmt_str += '%%-%ds' % field_len
+        if i != len(field_lens)-1:
+            fmt_str += '\t'
```
boegel: same here:

```python
fmt_str = ''.join(['%%-%ds' % fl for fl in field_lens])
fmt_str += '\t'
```

or even

```python
fmt_str = ''.join(['%%-%ds' % fl for fl in field_lens]) + '\t' 
```

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+@author: Ewan Higgs (Universiteit Gent)
+"""
+
+from itertools import chain
+
+def _mk_fmt_str(fields):
+    """Calculate the format string for printing the template parameters."""
+    field_lens = []
+    for field in fields:
+        field_lens.append(max(map(len, field)))
```
boegel:
```python
field_lens = [max(map(len, field)) for field in fields]
```

hod/table.py
```diff
@@ -0,0 +1,89 @@
...
+# You should have received a copy of the GNU General Public License
+# along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
+# #
+"""
+Utility functions for printing text in tables.
+
+@author: Ewan Higgs (Universiteit Gent)
+"""
+
+from itertools import chain
```
boegel: no used, please drop

test/unit/subcommands/test_subcommands_helptemplate.py
```diff
@@ -36,8 +37,11 @@
 class TestHelpTemplateApplication(unittest.TestCase):
     def test_run(self):
         app = HelpTemplateSubCommand()
-        with patch('hod.config.template.mklocalworkdir', return_value='localworkdir'):
-            app.run([])
+        environ = dict(PBS_DEFAULT='pbsmaster')
+        environ.update(os.environ)
```
boegel: what if `$PBS_DEFAULT` is defined in the environment?

```python
test_environ = copy.deepcopy(os.environ)
test_environ['PBS_DEFAULT'] = 'pbsmaster'
with patch('os.environ', test_environ):
```

boegel:
@ehiggs: reviewed, ping me when you're done with the remarks

hod/subcommands/helptemplate.py
ehiggs: Someone deep down wants `PBS_JOBID` defined. So I define it.

test/unit/subcommands/test_subcommands_helptemplate.py
ehiggs: That's fine. It needs to be defined.

hod/table.py
ehiggs: That's javadoc notation. `reST` uses `:param` but we don't even use that.  What I have is somewhat similar to the [Google styleguide suggestion](https://google-styleguide.googlecode.com/svn/trunk/pyguide.html) but the docstrings aren't really consistent across the project (it's an end user tool, not a library).

hod/table.py
ehiggs: This has been addressed.

hod/subcommands/helptemplate.py
```diff
@@ -74,11 +75,15 @@ class HelpTemplateSubCommand(SubCommand):
 
     def run(self, args):
         """Run 'help-template' subcommand."""
-        optparser = HelpTemplateOptions(go_args=args, envvar_prefix=self.envvar_prefix, usage=self.usage_txt)
+        # We need PBS_JOBID defined in the environment to generate localworkdirs
+        if 'PBS_DEFAULT' not in os.environ:
+            os.environ['PBS_DEFAULT'] = 'pbs-master'
+        os.environ['PBS_JOBID'] = '123.%s' % os.environ['PBS_DEFAULT']
```
boegel: this is just wrong, especially without any form of logging...

where is `$PBS_JOBID` required exactly? deal with not having it defined there somehow, by using a fallback value if `$PBS_JOBID` is not defined, not here, this is a hack at best

hod/table.py
boegel: ok

test/unit/subcommands/test_subcommands_helptemplate.py
```diff
@@ -36,8 +37,11 @@
 class TestHelpTemplateApplication(unittest.TestCase):
     def test_run(self):
         app = HelpTemplateSubCommand()
-        with patch('hod.config.template.mklocalworkdir', return_value='localworkdir'):
-            app.run([])
+        test_environ = dict(PBS_DEFAULT='pbsmaster')
```
boegel: why is this suddenly required? is this related to the `$PBS_JOBID` hack?

boegel:
@ehiggs: rereviewed, the `$PBS_JOBID` hack is a blocker imho

test/unit/subcommands/test_subcommands_helptemplate.py
ehiggs: Yes.

boegel: OK, then I'd like to see it disappear too. ;-)

hod/subcommands/helptemplate.py
ehiggs: Something is required because the help-template command needs to print the localworkdir. The localworkdir now uses the jobid to group the underlying local work directories. But in the case of printing the job it, we don't have a job and hence no jobid.

boegel: OK, but then handle that where you need the job ID somehow, not by injecting a dummy value in the environment.

What if there are other places (in the future) that need to have/print the localworkdir without being in a job?

ehiggs: This is a special command for printing diagnostic information outside a job. In all other cases, it should be an error to try to figure out a localworkdir without a job id. There could be some refactoring that takes place, but that would just be to pull it from the environment and then thread the job id to all calls until it gets to the localworkdir.

test/unit/subcommands/test_subcommands_helptemplate.py
ehiggs: Me too. But it's running outside the normal job context.

boegel:
Still feel a bit queasy about the `$PBS_JOBID` hack, but at least it's implemented cleanly now.

Going in, thanks @ehiggs!

ehiggs:
Thanks!

PR #100, opened 2015-10-14
Hod clone

### LINKED ISSUE #97: hod clone  - clones a dist in a target directory
We would like to use zipped eggs for deployment since it means shared file systems are abused less on startup (meaning we will have better interactivity). However, if the `etc` dir is zipped it makes it harder for people to make new hodconf distributions based on an existing dist. Therefore, we want a `hod clone` command as follows:

`hod clone --dist=<some-dist> --output-dir=<some-output-dir>`

`--dist` and `--output-dir` are non positional required arguments. This will copy the contents of the `<some-dist>` to `<some-output-dir>`.

ehiggs:
This PR should fix #97.

### DIFF
--- a/docs/Command_line_interface.rst
+++ b/docs/Command_line_interface.rst
@@ -27,15 +27,16 @@ Running ``hod`` without arguments is equivalent to ``hod --help``, and results i
     hanythingondemand version 3.0.0 - Run services within an HPC cluster
     usage: hod <subcommand> [subcommand options]
     Available subcommands (one of these must be specified!):
-        create          Submit a job to spawn a cluster on a PBS job controller
         batch           Submit a job to spawn a cluster on a PBS job controller, run a job script, and tear down the cluster when it's done
-        list            List submitted/running clusters
-        dists           List the available distributions
-        help-template   Print the values of the configuration templates based on the current machine.
-        genconfig       Write hod configs to a directory for diagnostic purposes
-        connect         Connect to a hod cluster.
         clean           Remove stale cluster info.
-        clone           Write hod configs to a directory for edtting purposes.
+        clone           Write hod configs to a directory for editing purposes.
+        connect         Connect to a hod cluster.
+        create          Submit a job to spawn a cluster on a PBS job controller
+        dists           List the available distributions
+        genconfig       Write hod configs to a directory for diagnostic purposes
+        help-template   Print the values of the configuration templates based on the current machine.
+        list            List submitted/running clusters
+        relabel         Change the label of an existing job.
 
 .. _cmdline_hod_options:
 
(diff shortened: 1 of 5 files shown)


boegel:
#100 \o/

ehiggs:
:100:!

docs/Command_line_interface.rst
```diff
@@ -35,6 +35,7 @@ Running ``hod`` without arguments is equivalent to ``hod --help``, and results i
         genconfig       Write hod configs to a directory for diagnostic purposes
         connect         Connect to a hod cluster.
         clean           Remove stale cluster info.
+        clone           Write hod configs to a directory for edtting purposes.
```
boegel: shouldn't we list these alphabetically?

any other order is arbitrary/confusing?

hod/main.py
```diff
@@ -45,6 +46,7 @@
     connect.ConnectSubCommand,
     clean.CleanSubCommand,
     relabel.RelabelSubCommand,
+    clone.CloneSubCommand,
```
boegel: let's sort these alphabetically...

hod/main.py
```diff
@@ -32,7 +32,8 @@
 import sys
 
 import hod
-from hod.subcommands import batch, connect, clean, create, dists, genconfig, helptemplate, relabel, listcmd
+from hod.subcommands import (batch, connect, clean, create, dists, genconfig,
+        helptemplate, relabel, listcmd, clone)
```
boegel: your alphabet is broken here with the last two (and `connect`) ;)

hod/subcommands/clone.py
```diff
@@ -0,0 +1,84 @@
...
+    HELP = "Write hod configs to a directory for editing purposes."
+
+    def run(self, args):
+        """Run 'genconfig' subcommand."""
+        optparser = CloneOptions(go_args=args, usage=self.usage_txt)
+        if len(optparser.args) < 3:
+            sys.stderr.write(self.usage())
+            sys.exit(1)
+
+        dist = optparser.args[1]
```
boegel: hmm, shouldn't we let people use `--dist` here too, like they do with `batch` and `create`?

hod/subcommands/clone.py
```diff
@@ -0,0 +1,84 @@
...
+
+class CloneSubCommand(SubCommand):
+    """Implementation of 'clone' subcommand."""
+
+    CMD = 'clone'
+    EXAMPLE = "<dist-to-copy> <output-directory>"
+    HELP = "Write hod configs to a directory for editing purposes."
+
+    def run(self, args):
+        """Run 'genconfig' subcommand."""
```
boegel: not genconfig

hod/subcommands/clone.py
```diff
@@ -0,0 +1,84 @@
...
+    """Option parser for 'clone' subcommand."""
+    VERSION = HOD_VERSION
+    ALLOPTSMANDATORY = False # let us use optionless arguments.
+
+
+class CloneSubCommand(SubCommand):
+    """Implementation of 'clone' subcommand."""
+
+    CMD = 'clone'
+    EXAMPLE = "<dist-to-copy> <output-directory>"
```
boegel: `--dist`?

hod/subcommands/clone.py
```diff
@@ -0,0 +1,84 @@
...
+        optparser = CloneOptions(go_args=args, usage=self.usage_txt)
+        if len(optparser.args) < 3:
+            sys.stderr.write(self.usage())
+            sys.exit(1)
+
+        dist = optparser.args[1]
+        output_dir = optparser.args[2]
+
+        if os.path.exists(output_dir):
+            sys.stderr.write('Error: Output directory exists.\n')
```
boegel: include output dir in error message

hod/subcommands/clone.py
```diff
@@ -0,0 +1,84 @@
...
+        try:
+            dists_dir = hcc.resolve_dists_dir()
+            src = os.path.join(dists_dir, dist)
+            if not os.path.exists(src):
+                sys.stderr.write('Error: Input dist not found: "%s".\n' % dist)
+                sys.exit(1)
+            shutil.copytree(src, output_dir)
+            return 0
+        except Exception as e:
+            _log.error("Failed to copy dist: %s", str(e))
```
boegel: hmm, does this work with the formatting?

we usually do `_log.error("Failed to copy dist: %s" % err)`

boegel:
@ehiggs: reviewed, small remarks

docs/Command_line_interface.rst
ehiggs: The order is based on the amount of use we anticipate. So the most important and used functions are at the top. If you think it should change (e.g. genconfig and helptemplate should be lower) then we can file a new issue.

boegel: I think we've come to a point where we have too many subcommands to order them based on amount of use we anticipate. So, the only other alternative that makes sense is to just sort alphabetically (easier on the eyes too imho).

hod/subcommands/clone.py
ehiggs: We use positional arguments like `cp`, `mv`, `ln`, `scp`, `rsync`, etc.

hod/subcommands/clone.py
ehiggs: Yeah it works. I don't even need the `str` call apparently.

```python
In [2]: from vsc.utils import fancylogger
In [3]: _log = fancylogger.getLogger(fname=False)
In [4]: _log.error('hello darkness my old %s', 'friend')
2015-10-14 11:20:47,175 ERROR      ipython         MainThread  hello darkness my old friend
In [3]: _log.error('hello darkness my old %s', str(ValueError('friend')))
2015-10-14 11:22:28,992 ERROR      ipython         MainThread  hello darkness my old friend
In [4]: _log.error('hello darkness my old %s', ValueError('friend'))
2015-10-14 11:22:35,428 ERROR      ipython         MainThread  hello darkness my old friend
```

hod/subcommands/clone.py
ehiggs: Discussed off PR. We use like `cp`, `mv`, etc here. `hod connect` and `hod relabel` also use positional arguments.

boegel: ok, less typing: good

boegel:
Good to go, thanks @ehiggs!

PR #108, opened 2015-10-20
add vsc-install as dep for testenv in tox.ini

boegel:
Adding vsc-install as well is required because it was split off from vsc-base, see https://github.com/hpcugent/vsc-base/pull/204

@ehiggs: please review


boegel:
PR tester fails on this, but due to a chicken-egg situation? @ehiggs?

Issue #121, opened 2015-10-28
$PBS_JOBID must be defined to create a localworkdir.


molden:
This relates to issue #92 . In my installation this is not working quite well yet. The PBS_JOBID is not known if I use more than 1 node.

ehiggs:
Fixed and released! Please use hod-3.0.1

PR #123, opened 2015-10-29
Send PBS_JOBID to the slaves so they can figure out the localworkdir.

ehiggs:
Intended to solve #121.


stdweird:
why is this needed? (why isn't pbs_jobid set in the job env of the slaves? sounds like a pbs bug...)

boegel:
@stdweird: doesn't `mympirun` also support specifying which environment variables should be passed down to the slaves, with a wildcard even, like `$PBS*`?

stdweird:
@boegel yes, `mympirun --variablesprefix=PBS` should do that.

ehiggs:
Indeed PBS_JOBID shows up in the slave. I have the following example script:

`env.sh`:
```bash
$ cat env.sh
#!/bin/bash

module load vsc-mympirun
module load OpenMPI
module load mpi4py
module load vsc-base
cd $PBS_O_WORKDIR
mympirun --hybrid 2 --output=${PBS_JOBID}.log $HOME/jobs/env.py
```

`env.py`:
```python
#!/usr/bin/env python

from mpi4py import MPI
import os
from vsc.utils import fancylogger

_log = fancylogger.getLogger(fname=False)

MASTERRANK = 0

if MPI.COMM_WORLD.rank == MASTERRANK:
    _log.warn('master: %s' % str(os.environ))
else:
    _log.warn('slave: %s' % str(os.environ))
```

And the `PBS_JOBID` is in the slave log. I'm going to close this and continue investigating.

PR #124, opened 2015-10-30
Make sure PBS prefix variables are sent to slaves.

ehiggs:
Fix for #121.


stdweird:
why is this added? it looks like a bug in torque which should be fixed instead worked around.

ehiggs:
@stdweird When I make a test script, the PBS prefix variables make it to the slaves, so I don't think it's Torque.

stdweird:
i don't understand how the variables could not have been there in the first place

PR #130, opened 2015-11-09
Use the env script we use on hod connect to set env for hod batch.

### LINKED ISSUE #128: hod batch uses LocalJobRunner
The `HADOOP_CONF_DIR` isn't loaded into the environment when the `hod batch` script job is run so it defaults to using the LocalJobRunner. This isn't what we want. The script should instead load the same environment script as `hod connect`.

ehiggs:
Fixes #128 This fixes a bug where hod used the local job runner which gives the appearance
of working; but was absolutely not what we wanted.

Also, I move `hod.hodproc.load_hod_config` to `hod.config.config` to prevent a cyclic dependency between `hod.hodproc` and `hod.cluster`.

### DIFF
--- a/hod/hodproc.py
+++ b/hod/hodproc.py
@@ -142,7 +142,7 @@ class ConfiguredMaster(MpiService):
             script = self.options.script
             script_stdout, script_stderr = _script_output_paths(script, label)
             redirection = ' > %s 2> %s' % (script_stdout, script_stderr)
-            start_script = env_script + '; ' + script + redirection + '; qdel $PBS_JOBID'
+            start_script = env_script + ' && ' + script + redirection + '; qdel $PBS_JOBID'
             self.log.debug('Adding script Task: %s', start_script)
             # TODO: How can we test this?
             config = ConfigOpts(script, RUNS_ON_MASTER, '', start_script, '', master_env, resolver, timeout=NO_TIMEOUT)


hod/cluster.py
```diff
@@ -46,12 +46,13 @@
 # make sure session is properly set up (e.g., that 'module' command is defined)
 source /etc/profile
 
+module load %(modules)s
```
boegel: why did you move this? just wondering if there's a particular reason, there's no problem with this change

hod/hodproc.py
```diff
@@ -146,10 +137,13 @@ def distribution(self, *master_template_args, **kwargs):
...
 
         if hasattr(self.options, 'script') and self.options.script is not None:
+            label = self.options.label
+            env_script = 'source ' + hc.cluster_env_file(label)
             script = self.options.script
-            script_stdout, script_stderr = _script_output_paths(script, self.options.label)
+            script_stdout, script_stderr = _script_output_paths(script, label)
             redirection = ' > %s 2> %s' % (script_stdout, script_stderr)
-            start_script = script + redirection + '; qdel $PBS_JOBID'
+            start_script = env_script + '; ' + script + redirection + '; qdel $PBS_JOBID'
```
boegel: do you want `source <envscript>; <script>` or `source <envscript> && <script>`?
if the `source` fails (for whatever reason), you probably don't want to start running the script?

also, should we try and make sure the exit code of the script being one trickles up and is used as exit code of the job being submitted? doesn't need to be done in this PR, just open an issue/PR for it if you think it makes sense

boegel:
@ehiggs: no way to test this to make sure it doesn't get broken again?

hod/cluster.py
ehiggs: Older Hadoop easyconfigs (e.g. Hadoop 2.3.0 which is installed at our site but not in easybuild's develop branch) set the `HADOOP_CONF_DIR` so setting it in the script and then loading the module breaks batch.

hod/hodproc.py
ehiggs: >do you want source <envscript>; <script> or source <envscript> && <script>?

Good idea.

ehiggs:
@boegel re testing: #132

Issue #129, opened 2015-11-09
Performance tune Spark


ehiggs:
By default, Spark uses something silly like only 500MB per executor and only three executors. We should add a Spark autogen script to calculate how many executors there should be and how much memory they should get.

ehiggs:
#131 has what appears to be correct parameters for Spark, and in the logs, I see that Spark is asking for the appropriate resources, but for some reason Yarn is not releasing them. There are two issues people are running into:

1. We ask for `spark.executor.instances 2` on a single node, but Yarn only grants one container. If we are on multiple nodes and ask for 5, then we only get 2 (one per node). This limits the cores and memory in use.

2. Some users in a classroom setting have their settings ignored completely and end up with the default Spark memory settings. i.e. 

```
spark.executor.cores 8
spark.executor.instances 2
spark.executor.memory 18432M
``` 

Results in 1 executor  with 512M.

ehiggs:
Executor stdout:

```
2015-12-02 17:06:26,103 INFO  [main] yarn.YarnRMClient (Logging.scala:logInfo(59)) - Registering the ApplicationMaster
2015-12-02 17:06:26,190 INFO  [main] yarn.YarnAllocator (Logging.scala:logInfo(59)) - Will request 2 executor containe
rs, each with 5 cores and 20275 MB memory including 1843 MB overhead
2015-12-02 17:06:26,198 INFO  [main] yarn.YarnAllocator (Logging.scala:logInfo(59)) - Container request (host: Any, ca
pability: <memory:20275, vCores:5>)
2015-12-02 17:06:26,199 INFO  [main] yarn.YarnAllocator (Logging.scala:logInfo(59)) - Container request (host: Any, ca
pability: <memory:20275, vCores:5>)
2015-12-02 17:06:26,232 INFO  [main] yarn.ApplicationMaster (Logging.scala:logInfo(59)) - Started progress reporter th
read with (heartbeat : 3000, initial allocation : 200) intervals
2015-12-02 17:06:27,645 INFO  [Reporter] impl.AMRMClientImpl (AMRMClientImpl.java:populateNMTokens(361)) - Received ne
w token for : node2159.delcatty.data:50680
2015-12-02 17:06:27,650 INFO  [Reporter] yarn.YarnAllocator (Logging.scala:logInfo(59)) - Launching container containe
r_1449072110115_0001_01_000002 for on host node2159.delcatty.data
[...]

PR #131, opened 2015-11-09
Spark tuning

ehiggs:
Spark defaults don't 'use everything'. Instead they are very crippled on Yarn. So we need to autogen the configs. This is a WIP to tune the defaults.

Addresses #129.

### DIFF
--- a/etc/hod/IPython-notebook-3.2.1/ipython.conf
+++ b/etc/hod/IPython-notebook-3.2.1/ipython.conf
@@ -11,3 +11,7 @@ ExecStop=
 [Environment]
 HADOOP_CONF_DIR=$localworkdir/conf
 HADOOP_LOG_DIR=$localworkdir/log
+SPARK_CONF_DIR=$localworkdir/conf
+SPARK_LOG_DIR=$localworkdir/log
+SPARK_LOG_DIR=$localworkdir/log
+SPARK_PID_DIR=$localworkdir/pid
--- a/etc/hod/IPython-notebook-3.2.1/ipython.conf
+++ b/etc/hod/IPython-notebook-3.2.1/ipython.conf
@@ -13,5 +13,4 @@ HADOOP_CONF_DIR=$localworkdir/conf
 HADOOP_LOG_DIR=$localworkdir/log
 SPARK_CONF_DIR=$localworkdir/conf
 SPARK_LOG_DIR=$localworkdir/log
-SPARK_LOG_DIR=$localworkdir/log
 SPARK_PID_DIR=$localworkdir/pid
--- a/hod/cluster.py
+++ b/hod/cluster.py
@@ -36,7 +36,7 @@ from collections import namedtuple
 from vsc.utils import fancylogger
 
 from hod.config.config import resolve_config_paths
-import hod.hodproc as hh
+import hod.config.config as hc
 
 
 _log = fancylogger.getLogger(fname=False)
@@ -46,12 +46,13 @@ CLUSTER_ENV_TEMPLATE = """
 # make sure session is properly set up (e.g., that 'module' command is defined)
 source /etc/profile
 
+module load %(modules)s
+
 # set up environment
 export HADOOP_CONF_DIR='%(hadoop_conf_dir)s'
 export HBASE_CONF_DIR='%(hadoop_conf_dir)s'
 export HOD_LOCALWORKDIR='%(hod_localworkdir)s'
 # TODO: HADOOP_LOG_DIR?
-module load %(modules)s
 
 echo "Welcome to your hanythingondemand cluster (label: %(label)s)"
 echo
@@ -214,7 +215,7 @@ def gen_cluster_info(label, options):
     """Generate cluster info as a dict, intended to use as template values for CLUSTER_ENV_TEMPLATE."""
     # list of modules that should be loaded: modules for selected service + extra modules specified via --modules
     config_path = resolve_config_paths(options.hodconf, options.dist)
-    hodconf = hh.load_hod_config(config_path, options.workdir, options.modules)
+    hodconf = hc.load_hod_config(config_path, options.workdir, options.modules)
     cluster_info = {
         'hadoop_conf_dir': hodconf.configdir,
         'hod_localworkdir': hodconf.localworkdir,
--- a/hod/config/autogen/ipython_notebook.py
+++ b/hod/config/autogen/ipython_notebook.py
@@ -35,6 +35,9 @@ def spark_defaults(_, node_info):
     '''
     Generate spark defaults so spark uses all the resources that yarn is able to
     provide.
+
+    Defaults here are based on Cloudera's recommendations here: 
+    http://blog.cloudera.com/blog/2015/03/how-to-tune-your-apache-spark-jobs-part-2/
     '''
     memory_defaults = hcah.memory_defaults(node_info)
     num_nodes = node_info['num_nodes']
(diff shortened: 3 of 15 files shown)


etc/hod/IPython-notebook-3.2.1/ipython.conf
```diff
@@ -11,3 +11,7 @@ ExecStop=
 [Environment]
 HADOOP_CONF_DIR=$localworkdir/conf
 HADOOP_LOG_DIR=$localworkdir/log
+SPARK_CONF_DIR=$localworkdir/conf
+SPARK_LOG_DIR=$localworkdir/log
+SPARK_LOG_DIR=$localworkdir/log
```
boegel: just once is fine?

hod/cluster.py
```diff
@@ -46,12 +46,13 @@
 # make sure session is properly set up (e.g., that 'module' command is defined)
 source /etc/profile
 
+module load %(modules)s
```
boegel: same stuff as in #130? so that one should be merged first

hod/config/autogen/ipython_notebook.py
```diff
@@ -28,32 +28,28 @@
...
-    template = string.Template(template)
-    # IPython.lib.passwd('hanythingondemand')
-    # 'sha1:906f33be372a:2c33c548645189cfa7a37ea0b77ccfb65852ba28'
-    passwd_hash = '906f33be372a:2c33c548645189cfa7a37ea0b77ccfb65852ba28'
-    file_contents = template.substitute(hashed_password=passwd_hash, workdir=workdir)
-    return file_contents
+    memory_defaults = hcah.memory_defaults(node_info)
+    num_nodes = node_info['num_nodes']
+    instances_per_node = 3
+    # -1 because we want one less executor instance on the application master
```
boegel: there's the -1?

boegel: ah, below, sorry

hod/config/autogen/ipython_notebook.py
```diff
@@ -28,32 +28,28 @@
...
+    instances = (num_nodes * instances_per_node) - 1
+    cores = (node_info['cores'] / instances_per_node)
+    memory = hcac.round_mb(memory_defaults.available_memory / instances_per_node)
+    dflts = {
+        'spark.executor.instances': instances,
+        'spark.executor.cores': cores,
+        'spark.executor.memory':  str(memory) + 'M',
+        'spark.local.dir': '$localworkdir'
+    }
+    return dflts
```
boegel: why obfuscate this? just use `defaults`?

hod/config/autogen/ipython_notebook.py
```diff
@@ -28,32 +28,28 @@
...
-"""
-    template = string.Template(template)
-    # IPython.lib.passwd('hanythingondemand')
-    # 'sha1:906f33be372a:2c33c548645189cfa7a37ea0b77ccfb65852ba28'
-    passwd_hash = '906f33be372a:2c33c548645189cfa7a37ea0b77ccfb65852ba28'
-    file_contents = template.substitute(hashed_password=passwd_hash, workdir=workdir)
-    return file_contents
+    memory_defaults = hcah.memory_defaults(node_info)
+    num_nodes = node_info['num_nodes']
+    instances_per_node = 3
```
boegel: why 3? totally unclear where that's coming from

hod/config/autogen/ipython_notebook.py
```diff
@@ -64,7 +60,15 @@ def autogen_config(workdir, node_info):
...
     '''
     cfg2fn = {
-        'ipython_notebook.config.py': ipython_notebook_config,
+    }
+    cfg2fn = {
+        # Pulled in from Hadoop
+        'core-site.xml': hcah.core_site_xml_defaults,
+        'mapred-site.xml': hcah.mapred_site_xml_defaults,
+        'yarn-site.xml': hcah.yarn_site_xml_defaults,
+        'capacity-scheduler.xml': hcah.capacity_scheduler_xml_defaults,
```
boegel: order alphabetically pls

hod/config/config.py
```diff
@@ -455,7 +455,10 @@ def service_config_fn(policy_path):
     module_name = '.'.join(policy_path_list[:-1])
     parent_pkg = '.'.join(policy_path_list[:-2])
     fn = policy_path_list[-1]
-    module = __import__(module_name, fromlist=[parent_pkg])
+    try:
+        module = __import__(module_name, fromlist=[parent_pkg])
+    except ImportError as err:
+        _log.error('Could not import module "%s" from "%s": %s', module_name, parent_pkg, err)
```
boegel: only log? this looks pretty fatal to me?

hod/config/config.py
```diff
@@ -506,3 +509,14 @@ def resolve_config_paths(config, dist):
...
     return path
+
+def load_hod_config(filenames, workdir, modules):
+    '''
+    Load the manifest config (hod.conf) files.
+    '''
+    m_config_filenames = parse_comma_delim_list(filenames)
+    _log.info('Loading "%s" manifest config', m_config_filenames)
+    m_config = PreServiceConfigOpts.from_file_list(m_config_filenames,
+            workdir=workdir, modules=modules)
```
boegel: why break this line? no need to (harder to read imho)

hod/config/writer/_hadoop.py
```diff
@@ -53,6 +53,8 @@ def _write_masters(outfile, options, template_resolver):
         output += '%s\n' % (value)
     return output
  
+def _write_text_file(outfile, options, template_resolver):
+    return options
```
boegel: no docstring, function name doesn't cover what it actually does, etc.

but, this is simply not finished yet, right? I better stop reviewing this then. ;-)

hod/config/writer/common.py
```diff
@@ -11,3 +11,21 @@ def kv2xml(name, value):
...
+
+    property1 value1
+    property2 value2
+
+    "spark-defaults.conf" is one such example:
+    https://spark.apache.org/docs/latest/configuration.html#dynamically-loading-spark-properties
+    """
+    output = ""
+
+    for k, v in sorted(options.items()):
```
boegel: no single-letter variables outside of list comprehensions please

hod/config/writer/common.py
```diff
@@ -11,3 +11,21 @@ def kv2xml(name, value):
...
+        name = template_resolver(k)
+        value = template_resolver(v)
+        output += '%s %s\n' % (name, value)
+    return output
```
boegel: This whole function is something like:

```python
return '\n'.join(['%s %s' % (template_resolver(k), template_resolver(v)) for k, v in sorted(options.items())])
```

hod/config/autogen/ipython_notebook.py
ehiggs: Spark defaults are atrocious. They use like 2 executors with each only like 1GB memory apiece. And if you want to scale it up using `spark.dynamicAllocation.enabled = true` then it requires more services to run as described [here](https://spark.apache.org/docs/1.2.0/job-scheduling.html#dynamic-resource-allocation).

hod/config/autogen/ipython_notebook.py
ehiggs: I pulled it from [here](http://blog.cloudera.com/blog/2015/03/how-to-tune-your-apache-spark-jobs-part-2/). Sadly they don't seem to do a good job explaining where they get this 'three executors per node' from. In any event, I'll add documentation about where I got the info from.

hod/config/writer/common.py
ehiggs: Yes. And you could even do `tr = template_resolver` to save more space. It's still a bit challenging to read.

PR #135, opened 2015-12-03
Yarn vcores

ehiggs:
Yarn has a hard coded default for `DEFAULT_NM_VCORES=8;` This affects the ability of Spark to get executors spawned and hence it leaves a lot of performance on the table. This fix fixes it.

WIP because I would like to also add something for single node 'clusters' because I currently leave space in the memory allocation for the Spark Driver but I don't think it would need 1/3 of the system's memory to drive just two executors. Maybe 512M is enough to work w/ 3 executors on the single node.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,6 @@
 # 3.0.3
 * Autogenerate Yarn vcore settings. Otherwise the hard coded default of '8' is
   chosen, which is not ideal.
-* Roll IPython, matplotlib, etc, modules to use intel-2015b toolchain.
 
 # 3.0.2
 * Fix for `hod batch` so it has the same environment as a shell started with
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# 3.0.3
+* Autogenerate Yarn vcore settings. Otherwise the hard coded default of '8' is
+  chosen, which is not ideal.
+* Roll IPython, matplotlib, etc, modules to use intel-2015b toolchain.
+
 # 3.0.2
 * Fix for `hod batch` so it has the same environment as a shell started with
   `hod connect`. This is required to use YARN instead of Hadoop LocalJobRunner.
--- a/etc/hod/IPython-notebook-3.2.1/hod.conf
+++ b/etc/hod/IPython-notebook-3.2.1/hod.conf
@@ -4,7 +4,7 @@
 version=1
 
 [Config]
-modules=Hadoop/2.6.0-cdh5.4.5-native,Spark/1.5.0,IPython/3.2.1-intel-2015a-Python-2.7.10,matplotlib/1.4.3-intel-2015a-Python-2.7.10
+modules=Hadoop/2.6.0-cdh5.4.5-native,Spark/1.5.0,IPython/3.2.1-intel-2015b-Python-2.7.10,matplotlib/1.4.3-intel-2015b-Python-2.7.10
 master_env=HADOOP_HOME,EBROOTHADOOP,JAVA_HOME,PYTHONPATH
 services=resourcemanager.conf,nodemanager.conf,ipython.conf,screen.conf
 config_writer=hod.config.writer.hadoop_xml
--- a/etc/hod/IPython-notebook-3.2.1/hod.conf
+++ b/etc/hod/IPython-notebook-3.2.1/hod.conf
@@ -4,6 +4,8 @@
 version=1
 
 [Config]
+# The start-notebook.sh requires IPython, matplotlib, and Spark to be loaded.
+# These are set here:
 modules=Hadoop/2.6.0-cdh5.4.5-native,Spark/1.5.0,IPython/3.2.1-intel-2015a-Python-2.7.10,matplotlib/1.4.3-intel-2015a-Python-2.7.10
 master_env=HADOOP_HOME,EBROOTHADOOP,JAVA_HOME,PYTHONPATH
 services=resourcemanager.conf,nodemanager.conf,ipython.conf,screen.conf
--- a/etc/hod/IPython-notebook-3.2.1/hod.conf
+++ b/etc/hod/IPython-notebook-3.2.1/hod.conf
@@ -4,7 +4,7 @@
 version=1
 
 [Config]
-modules=Hadoop/2.6.0-cdh5.4.5-native,Spark/1.5.0,IPython/3.2.1-intel-2015b-Python-2.7.10,matplotlib/1.4.3-intel-2015b-Python-2.7.10
+modules=Hadoop/2.6.0-cdh5.4.5-native,Spark/1.5.0,IPython/3.2.1-intel-2015a-Python-2.7.10,matplotlib/1.4.3-intel-2015a-Python-2.7.10
 master_env=HADOOP_HOME,EBROOTHADOOP,JAVA_HOME,PYTHONPATH
 services=resourcemanager.conf,nodemanager.conf,ipython.conf,screen.conf
 config_writer=hod.config.writer.hadoop_xml
--- a/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
+++ b/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
@@ -3,9 +3,8 @@
 # exit as soon as an error occurs
 set -e
 
-#module load IPython/3.2.1-intel-2015a-Python-2.7.10
-#module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
-#module load Spark/1.5.0
+# This requires IPython, matplotlib, and Spark to be loaded. These are set in
+# the hod.conf file.
 
 ipython profile create nbserver
 
--- a/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
+++ b/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
@@ -3,8 +3,8 @@
 # exit as soon as an error occurs
 set -e
 
-module load IPython/3.2.1-intel-2015a-Python-2.7.10
-module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
+module load IPython/3.2.1-intel-2015b-Python-2.7.10
+module load matplotlib/1.4.3-intel-2015b-Python-2.7.10
 module load Spark/1.5.0
 
 ipython profile create nbserver
--- a/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
+++ b/etc/hod/IPython-notebook-3.2.1/start-notebook.sh
@@ -3,9 +3,9 @@
 # exit as soon as an error occurs
 set -e
 
-module load IPython/3.2.1-intel-2015b-Python-2.7.10
-module load matplotlib/1.4.3-intel-2015b-Python-2.7.10
-module load Spark/1.5.0
+#module load IPython/3.2.1-intel-2015a-Python-2.7.10
+#module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
+#module load Spark/1.5.0
 
 ipython profile create nbserver
 
--- a/hod/config/autogen/hadoop.py
+++ b/hod/config/autogen/hadoop.py
@@ -122,7 +122,7 @@ def yarn_site_xml_defaults(workdir, node_info):
     Default entries for the yarn-site.xml config file.
     '''
     mem_dflts = memory_defaults(node_info)
-
+    ncores = node_info['cores']
     max_alloc = round_mb(mem_dflts.ram_per_container * mem_dflts.num_containers)
     min_alloc = round_mb(mem_dflts.ram_per_container)
     dflts = {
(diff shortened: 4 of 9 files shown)


etc/hod/IPython-notebook-3.2.1/hod.conf
```diff
@@ -4,7 +4,7 @@
 version=1
 
 [Config]
-modules=Hadoop/2.6.0-cdh5.4.5-native,Spark/1.5.0,IPython/3.2.1-intel-2015a-Python-2.7.10,matplotlib/1.4.3-intel-2015a-Python-2.7.10
+modules=Hadoop/2.6.0-cdh5.4.5-native,Spark/1.5.0,IPython/3.2.1-intel-2015b-Python-2.7.10,matplotlib/1.4.3-intel-2015b-Python-2.7.10
```
boegel: what's your motivation for this?

if there's no compelling reason for it, then don't

we should move forward to 2015b with IPython 4.x (i.e., a new dist)

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
```diff
@@ -3,8 +3,8 @@
 # exit as soon as an error occurs
 set -e
 
-module load IPython/3.2.1-intel-2015a-Python-2.7.10
-module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
+module load IPython/3.2.1-intel-2015b-Python-2.7.10
+module load matplotlib/1.4.3-intel-2015b-Python-2.7.10
```
boegel: I wouldn't...

also, looking into a single location to list modules would be nice (open an issue?)

hod/config/autogen/hadoop.py
```diff
@@ -139,6 +139,11 @@ def yarn_site_xml_defaults(workdir, node_info):
         'yarn.resourcemanager.webapp.https.address': '$masterhostaddress:8090',
         'yarn.resourcemanager.scheduler.class': 'org.apache.hadoop.yarn.server.resourcemanager.scheduler.capacity.CapacityScheduler',
         'yarn.scheduler.capacity.allocation.file': 'capacity-scheduler.xml',
+        'yarn.app.mapreduce.am.resource.cpu-vcores': '1',
```
boegel: what's this hardcoded `1` avout?

test/unit/config/autogen/test_autogen_hadoop.py
```diff
@@ -70,7 +70,7 @@ def test_yarn_site_xml_defaults(self):
                 cores=24, totalcores=24, usablecores=range(24), num_nodes=1,
                 memory=dict(meminfo=dict(memtotal=68719476736), ulimit='unlimited'))
         d = hca.yarn_site_xml_defaults('/', node)
-        self.assertEqual(len(d), 13)
+        self.assertEqual(len(d), 17)
```
boegel: can you give a more meaningful name to `d`, so it's somewhat clear what the `17` refers to?

test/unit/config/autogen/test_autogen_hadoop_on_lustre2.py
```diff
@@ -67,7 +67,7 @@ def test_yarn_site_xml_defaults(self):
                 cores=4, totalcores=24, usablecores=[0, 1, 2, 3], num_nodes=1,
                 memory=dict(meminfo=dict(memtotal=68719476736), ulimit='unlimited'))
         d = hca.yarn_site_xml_defaults('/', node)
-        self.assertEqual(len(d), 14)
+        self.assertEqual(len(d), 18)
```
boegel: same here, better name?

boegel:
@ehiggs: reviewed, let me know if it's ready to check again

etc/hod/IPython-notebook-3.2.1/hod.conf
ehiggs: We've installed packages for users using 2015b so if we continue to use 2015a here, there is a conflict.

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
ehiggs: Agree. It should have the modules loaded from the hod.conf, but iirc it didn't work correctly in testing. I should have fixed it then.

hod/config/autogen/hadoop.py
ehiggs: This setting says for yarn to run the mapreduce app with 1 vcore per container (unless otherwise overridden). This doesn't affect Spark and actually I think it won't really be taken into account for MapReduce because we are using the DominantResourceCalculator in the CapacityScheduler which only takes memory into account.

test/unit/config/autogen/test_autogen_hadoop.py
ehiggs: This is a test and it's a dict. If you read the code you see that it is asserting the length of 'd' is 17. 17 is the number of things in the dict w/ settings.

etc/hod/IPython-notebook-3.2.1/hod.conf
boegel: We can install those packages with 2015a to make them compatible with this IPython dist, I don't feel that's a good reason to redefine an existing dist.

If anything, it's an argument to include the toolchain used in the dist name...

hod/config/autogen/hadoop.py
boegel: can you add a comment above this line to clarify this in the code, preferably with a reference to the docs?

the other settings are fairly easy to understand imho

test/unit/config/autogen/test_autogen_hadoop.py
boegel: I know it's a test, but it's unclear. Anyway, if you feel changing the name isn't worth the trouble, fine.

It's just that it 'violates' our code style ("no single-letter variables outside of list comprehensions")

ehiggs:
Please check again.
I've reduced the number of cores / executor to 2 based on 'research;. It performs twice as fast in the test cases I've given it (PySpark notebook).

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
```diff
@@ -3,9 +3,9 @@
 # exit as soon as an error occurs
 set -e
 
-module load IPython/3.2.1-intel-2015a-Python-2.7.10
-module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
-module load Spark/1.5.0
+#module load IPython/3.2.1-intel-2015a-Python-2.7.10
+#module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
+#module load Spark/1.5.0
```
boegel: why drop these `module load`s?

NEWS.md
```diff
@@ -1,3 +1,8 @@
+# 3.0.3
+* Autogenerate Yarn vcore settings. Otherwise the hard coded default of '8' is
+  chosen, which is not ideal.
+* Roll IPython, matplotlib, etc, modules to use intel-2015b toolchain.
```
boegel: drop this item, we're not doing that anymore?

hod/config/autogen/ipython_notebook.py
```diff
@@ -41,16 +42,18 @@ def spark_defaults(_, node_info):
     '''
     memory_defaults = hcah.memory_defaults(node_info)
     num_nodes = node_info['num_nodes']
-    instances_per_node = 3
+    cores_per_executor = 2
+    instances_per_node = node_info['cores'] / cores_per_executor
```
boegel: what with single-core jobs?

```
>>> 1/2
0
```

shouldn't we have:

```python
instances_per_node = max(node_info['cores'] / cores_per_executor, 1)
```

hod/config/autogen/ipython_notebook.py
```diff
@@ -41,16 +42,18 @@ def spark_defaults(_, node_info):
     '''
     memory_defaults = hcah.memory_defaults(node_info)
     num_nodes = node_info['num_nodes']
-    instances_per_node = 3
+    cores_per_executor = 2
```
boegel: include a comment how you picked this, because we will questioning this again later

test/unit/config/autogen/test_ipython_notebook.py
```diff
@@ -39,6 +39,16 @@ def test_spark_defaults(self):
                     memory=dict(meminfo=dict(memtotal=68719476736), ulimit='unlimited'))
         dflts = hcip.spark_defaults(None, node)
         self.assertTrue(len(dflts), 3)
-        self.assertEqual(dflts['spark.executor.instances'], 17)
-        self.assertEqual(dflts['spark.executor.cores'], 5)
-        self.assertEqual(hcc.parse_memory(dflts['spark.executor.memory']), hcc.parse_memory('18G'))
+        self.assertEqual(dflts['spark.executor.instances'], 47) # 6*8 -1
```
boegel: clarify the comment: `# 6 nodes * 8 cores  - 1`

test/unit/config/autogen/test_ipython_notebook.py
```diff
@@ -39,6 +39,16 @@ def test_spark_defaults(self):
...
+        self.assertEqual(dflts['spark.executor.cores'], 2)
+        self.assertEqual(hcc.parse_memory(dflts['spark.executor.memory']), hcc.parse_memory('7G'))
+
+    def test_spark_defaults_single_node(self):
```
boegel: add a single-core test too?

boegel:
@ehiggs: rereviewed, can't merge this yet imho due to the possible issues with single-core jobs...

hod/config/autogen/ipython_notebook.py
ehiggs: Single core hadoop clusters is absurd but yes this is correct. Good catch.
On 7 Dec 2015 10:30 am, Kenneth Hoste <notifications@github.com> wrote:In hod/config/autogen/ipython_notebook.py:
> @@ -41,16 +42,18 @@ def spark_defaults(_, node_info):
>      '''
>      memory_defaults = hcah.memory_defaults(node_info)
>      num_nodes = node_info['num_nodes']
> -    instances_per_node = 3
> +    cores_per_executor = 2
> +    instances_per_node = node_info['cores'] / cores_per_executor

what with single-core jobs?

>>> 1/2
0


shouldn't we have:

instances_per_node = max(node_info['cores'] / cores_per_executor, 1)

test/unit/config/autogen/test_ipython_notebook.py
ehiggs: +1
On 7 Dec 2015 10:31 am, Kenneth Hoste <notifications@github.com> wrote:In test/unit/config/autogen/test_ipython_notebook.py:
> @@ -39,6 +39,16 @@ def test_spark_defaults(self):
>                      memory=dict(meminfo=dict(memtotal=68719476736), ulimit='unlimited'))
>          dflts = hcip.spark_defaults(None, node)
>          self.assertTrue(len(dflts), 3)
> -        self.assertEqual(dflts['spark.executor.instances'], 17)
> -        self.assertEqual(dflts['spark.executor.cores'], 5)
> -        self.assertEqual(hcc.parse_memory(dflts['spark.executor.memory']), hcc.parse_memory('18G'))
> +        self.assertEqual(dflts['spark.executor.instances'], 47) # 6*8 -1

clarify the comment: # 6 nodes * 8 cores  - 1

NEWS.md
ehiggs: +1
On 7 Dec 2015 10:28 am, Kenneth Hoste <notifications@github.com> wrote:In NEWS.md:
> @@ -1,3 +1,8 @@
> +# 3.0.3
> +* Autogenerate Yarn vcore settings. Otherwise the hard coded default of '8' is
> +  chosen, which is not ideal.
> +* Roll IPython, matplotlib, etc, modules to use intel-2015b toolchain.

drop this item, we're not doing that anymore?

etc/hod/IPython-notebook-3.2.1/start-notebook.sh
ehiggs: These are pulled in by the hod.conf. I'll remove them but add a comment that it depends on modules loaded in another file. I'll add the coresspinding comment in hod.conf.
On 7 Dec 2015 10:27 am, Kenneth Hoste <notifications@github.com> wrote:In etc/hod/IPython-notebook-3.2.1/start-notebook.sh:
> @@ -3,9 +3,9 @@
>  # exit as soon as an error occurs
>  set -e
>  
> -module load IPython/3.2.1-intel-2015a-Python-2.7.10
> -module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
> -module load Spark/1.5.0
> +#module load IPython/3.2.1-intel-2015a-Python-2.7.10
> +#module load matplotlib/1.4.3-intel-2015a-Python-2.7.10
> +#module load Spark/1.5.0

why drop these module loads?

boegel:
@ehiggs: looks good, how can I test this? You told me you didn't retest after your last change, so I'm kind of reluctant to merge it in (although all changes make sense to me now).

ehiggs:
This has been fairly well tested. We need to get this into the hands of users to break it further.

boegel:
Going in, thanks @ehiggs!

PR #139, opened 2015-12-08
Nolabel label

ehiggs:
When we don't set a label and the system generates one for us, we need to store it so the rest of the cluster knows what it is.

### DIFF
--- a/hod/local.py
+++ b/hod/local.py
@@ -84,6 +84,7 @@ def main(args):
             # if no label is specified, use job ID;
             # if $PBS_JOBID is not set, generate a random string (10 chars)
             label = os.getenv('PBS_JOBID', ''.join(random.choice(string.letters + string.digits) for _ in range(10)))
+            optparser.options.label = label
 
         _log.debug("Creating cluster info using label '%s'", label)
         cluster_info = gen_cluster_info(label, optparser.options)


boegel:
@ehiggs: please rebase on `master`...?

ehiggs:
@boegel Rebased. Please review.

boegel:
@ehiggs: what issue are you fixing with this exactly, how did you notice this?

I'm not sure that changing 'optparser.options' is the best way to go here, especially since we pass down 'label' in other places.

ehiggs:
I noticed this by running without the `--label` option with batch. It failed. 

`label` appears to be set to PBS_JOBID but then the only thing passed down to `ConfiguredMaster` is the `optpaster.options`. Where do you see it being passed down in other places?

ehiggs:
Any further questions/objections before merging this?

hod/local.py
```diff
@@ -84,6 +84,7 @@ def main(args):
             # if no label is specified, use job ID;
             # if $PBS_JOBID is not set, generate a random string (10 chars)
             label = os.getenv('PBS_JOBID', ''.join(random.choice(string.letters + string.digits) for _ in range(10)))
+            optparser.options.label = label
 
         _log.debug("Creating cluster info using label '%s'", label)
         cluster_info = gen_cluster_info(label, optparser.options)
```
boegel: we're passing down `label` here, even though it's in `optparser.options` too, which feels wrong

let me look into a small PR your way to do this slightly differently, changing the parsed options is just wrong imho

ehiggs: The parsed command line options are supposed to hold all the information to run `hod.local`. The exception is the label which can be omitted so that it is generated when the job is run (using `PBS_JOBID`). The only sane alternative, I guess, is to copy everything into a separate namedtuple and then pass that. But is that really better?

boegel: OK, it's going to be messy to do this 'properly', so let's merge this in as is.

I've opened #140 with an idea to clean up the way we're doing this now.

PR #141, opened 2015-12-11
Make spark use /tmp for local dir.

ehiggs:
Spark normally takes the yarn local dir. However, it does this only from the
local dir on the master node. This means all the spark executors on all the
nodes want to talk to the one hod localworkdir used by the master. This commit
changes this to use /tmp

Another solution would have been to set HOD_LOCALWORKDIR in the environments and
use {{HOD_LOCALWORKDIR}} and let spark's templating resolve it later.

### DIFF
--- a/hod/config/autogen/ipython_notebook.py
+++ b/hod/config/autogen/ipython_notebook.py
@@ -56,7 +56,7 @@ def spark_defaults(_, node_info):
         'spark.executor.cores': cores_per_executor,
         'spark.executor.instances': instances,
         'spark.executor.memory':  str(memory) + 'M',
-        'spark.local.dir': '$localworkdir',
+        'spark.local.dir': '/tmp',
     }
     return dflts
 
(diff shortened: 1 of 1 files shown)


hod/config/autogen/ipython_notebook.py
```diff
@@ -56,7 +56,7 @@ def spark_defaults(_, node_info):
         'spark.executor.cores': cores_per_executor,
         'spark.executor.instances': instances,
         'spark.executor.memory':  str(memory) + 'M',
-        'spark.local.dir': '$localworkdir',
+        'spark.local.dir': '/tmp',
```
boegel: use `tempfile.gettempdir()` ?

see https://docs.python.org/2/library/tempfile.html#tempfile.gettempdir

it will pick up on `$TMPDIR`, which is defined in PBS jobs as `/tmp/$PBS_JOBID`

boegel:
I suggest you test this first at length before we merge it?

ehiggs:
This has been tested now. It seems to work v. well. :rocket:

boegel:
Thanks @ehiggs.

Issue #22, opened 2015-04-19
time to usable screen is too long on multinode job


stdweird:
The time to screen session is very long, it's unclear what causes this

ehiggs:
In taking a look at this, I'm thinking of using a dependency graph to order the process starts. This will 

a.) give us a useful ordering or service starts, 

b.) if we have an ordering it will allow scripts to only start when services are up. 

c.) increase debugability since we can define the order that servies come up, we can test to make sure it happens in this order. Currently, they can come up in perhaps any order they like on all the machines.

When we have a dependency graph we can then push the screen session to start up earlier.

stdweird:
sure it would be nice, but for now, you know the order right? so i fail to see how this will help (if not only make things more complicated to debug).

ehiggs:
Plotting the log entries, the time has several jumps. These are each one minute long because of the way the mpiservice job loop runs, waiting for a minute each time through.

If we change the wait time on the loop, it gets much quicker on 2 nodes. In the following plot, the x axis is log message# (cropped to when the screen service has started). The y axis is seconds from the first log message.

![time-to-startup](https://cloud.githubusercontent.com/assets/28823/12356717/51cbc53a-bba4-11e5-93da-f0111394026a.png)

ehiggs:
Plotting the log entries, the time has several jumps. These are each one minute long because of the way the mpiservice job loop runs, waiting for a minute each time through.

If we change the wait time on the loop, it gets much quicker on 2 nodes. The following plot is 'line in the log file' on the x axis and is cropped to when the screen service has started. The y axis is seconds from the first log message.

![time-to-startup](https://cloud.githubusercontent.com/assets/28823/12356471/3219cc10-bba3-11e5-84ac-2de9665d0ac0.png)

PR #148, opened 2016-01-15
Reduce time to check processes down to 15 from 60.

ehiggs:
Seems to make our startup faster my a linear amount.

Addresses #22.

### DIFF
--- a/hod/mpiservice.py
+++ b/hod/mpiservice.py
@@ -194,7 +194,7 @@ def run_tasks(svc):
     """Make communicators for tasks and execute the work there"""
     # Based on initial dist, create the groups and communicators and map with work
     active_work = []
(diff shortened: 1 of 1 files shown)


boegel:
@ehiggs: what's up with the failing tests?

Also, would it hurt to make this even tighter, e.g. 5s?

boegel:
As discussed: can we use 5s during startup, and then switch back to 60s once the cluster is running?

boegel:
@ehiggs: what's up with the tests? We need them green to merge this.

ehiggs:
Jenkins, test this please.

ehiggs:
I don't know what's up with the test. Jenkins claims it timed out fetching the branch after 10 minutes.

PR #161, opened 2016-08-11
add support for --modulepaths to specify the location of custom modules

boegel:
Users can already support additional modules to be loaded via `--modules`, but these are expected to be available by default in whichever `$MODULEPATH` that gets defined.

With Lmod as a modules tool, specifying the modules to be loaded full path doesn't work.

By adding `--modulepaths` and running `module use` on each of them, users can specify additional locations for modules to be loaded.

### DIFF
--- a/hod/cluster.py
+++ b/hod/cluster.py
@@ -46,10 +46,7 @@ CLUSTER_ENV_TEMPLATE = """
 # make sure session is properly set up (e.g., that 'module' command is defined)
 source /etc/profile
 
-for modpath in %(modpaths)s;
-do
-    module use modpath
-done
+for modpath in %(modpaths)s; do module use $modpath; done
 
 module load %(modules)s
 
--- a/hod/cluster.py
+++ b/hod/cluster.py
@@ -217,7 +217,7 @@ def gen_cluster_info(label, options):
     """Generate cluster info as a dict, intended to use as template values for CLUSTER_ENV_TEMPLATE."""
     # list of modules that should be loaded: modules for selected service + extra modules specified via --modules
     config_path = resolve_config_paths(options.hodconf, options.dist)
-    hodconf = hc.load_hod_config(config_path, options.workdir, options.modules)
+    hodconf = hc.load_hod_config(config_path, options.workdir, options.modulepaths, options.modules)
     cluster_info = {
         'hadoop_conf_dir': hodconf.configdir,
         'hod_localworkdir': hodconf.localworkdir,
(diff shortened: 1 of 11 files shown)


hod/cluster.py
```diff
@@ -46,6 +46,11 @@
 # make sure session is properly set up (e.g., that 'module' command is defined)
 source /etc/profile
 
+for modpath in %(modpaths)s;
+do
+    module use modpath
```
stdweird: `$modpath`?

hod/cluster.py
```diff
@@ -46,6 +46,11 @@
 # make sure session is properly set up (e.g., that 'module' command is defined)
 source /etc/profile
 
+for modpath in %(modpaths)s;
+do
```
stdweird: move the do to previouos line if you end it with an `;`

hod/cluster.py
boegel: ugh, good catch (I was still verifying whether this actually works, clearly it won't :))

hod/local.py
```diff
@@ -64,6 +64,7 @@ def config_options(self):
         """Add general configuration options."""
         opts = copy.deepcopy(GENERAL_HOD_OPTIONS)
         opts.update({
+            'modulepaths': ("Extra paths to take into account for loading modules", 'string', 'store', None),
```
stdweird: copy/paste much?

boegel: some options only make sense for certain subcommands, while others don't (cfr. `--script`)

we're already using `GENERAL_HOD_OPTIONS` for options that are common for *all* subcommands, I don't think we should take it much further than that?

stdweird: sure, but why copy paste all the common ones, incl the lines above etc etc. and people say python is OO....

boegel: fixed, I added a `COMMON_HOD_CONFIG_OPTIONS` too and used it where , I don't think an OO approach would be a lot better

hod/subcommands/batch.py
```diff
@@ -69,9 +69,7 @@ def batch_options(self):
     def config_options(self):
         """Add general configuration options."""
         opts = copy.deepcopy(GENERAL_HOD_OPTIONS)
-        opts.update({
-            'modules': ("Extra modules to load in each service environment", 'string', 'store', None),
-        })
+        opts.update(COMMON_HOD_CONFIG_OPTIONS)
```
stdweird: thx, much better

boegel:
This is working as expected now, thanks for the review @stdweird!

PR #163, opened 2016-08-17
add dist for Jupyter notebook 5.1.0 w/ Spark 2.0.0

boegel:
tested with https://github.com/apache/spark/blob/master/examples/src/main/python/pi.py, seems to work fine

### DIFF
--- a/hod/__init__.py
+++ b/hod/__init__.py
@@ -27,7 +27,7 @@
 
 @author: Stijn De Weirdt (Ghent University)
 @author: Ewan Higgs (Ghent University)
-@author: Ewan Higgs (Ghent University)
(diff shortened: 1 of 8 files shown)


hod/__init__.py
```diff
@@ -23,10 +23,11 @@
 # along with hanythingondemand. If not, see <http://www.gnu.org/licenses/>.
 ##
 """
-Nothing here for now.
+'hod' namespace declaration, also specifies HOD version
 
 @author: Stijn De Weirdt (Ghent University)
 @author: Ewan Higgs (Ghent University)
+@author: Ewan Higgs (Ghent University)
```
stdweird: @boegel i guess this should be you

PR #191, opened 2017-02-16
pass string rather than Exception instance as argument to raiseException

boegel:
fix for this nasty traceback, which hides the actual problem (some `ImportError`):

```
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/apps/gent/SL6/sandybridge/software/hanythingondemand/3.2.0-cli/lib/python2.6/site-packages/hanythingondemand-3.2.0-py2.6.egg/hod/main.py", line 92, in <module>
  File "/apps/gent/SL6/sandybridge/software/hanythingondemand/3.2.0-cli/lib/python2.6/site-packages/hanythingondemand-3.2.0-py2.6.egg/hod/main.py", line 80, in main
  File "/apps/gent/SL6/sandybridge/software/hanythingondemand/3.2.0-cli/lib/python2.6/site-packages/hanythingondemand-3.2.0-py2.6.egg/hod/subcommands/listcmd.py", line 90, in run
  File "/usr/lib/python2.7/site-packages/vsc/utils/fancylogger.py", line 322, in raiseException
    message += " (%s)" % detail
TypeError: unsupported operand type(s) for +=: 'exceptions.ImportError' and 'str'
```

### DIFF
--- a/hod/__init__.py
+++ b/hod/__init__.py
@@ -30,4 +30,4 @@
 @author: Kenneth Hoste (Ghent University)
 """
 NAME = 'hanythingondemand'
-VERSION = '3.2.0'
+VERSION = '3.2.1'
--- a/hod/main.py
+++ b/hod/main.py
@@ -77,8 +77,7 @@ def main(args):
     """Parse options and run specified subcommand."""
     subcmd, args = init_subcmd(args)
     if subcmd is not None:
-        subcmd.run(args)
-        return 0
+        return subcmd.run(args)
 
     elif len([arg for arg in args if not arg.startswith('-')]) > 1:
         sys.stderr.write("ERROR: No known subcommand specified\n")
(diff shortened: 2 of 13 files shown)


stdweird:
rebase?
also 1st class copy/paste code....

boegel:
Actual problem was a broken `pbs_python` module that was used by HOD, because it was installed on SL6 where the system Python was 2.6, and was then being used on CentOS7:

```
libpython2.6.so.1.0: cannot open shared object file: No such file or directory; required module 'pbs' is not available
```

Reinstalling the `pbs_python` module fixed the problem.

@stdweird You're right that this could use a bit of cleanup, I'll look into that.

PR #193, opened 2017-05-04
fix undefined variable in mk_cluster_info_dict, report non-directory entries in cluster info dir

boegel:
fix for #192

PR #195, opened 2017-05-05
fix isdir check in known_cluster_labels, use full path to location to check

boegel:
bug that slipped in with #193 due to incomplete test...


itkovian:
LGTM, so merge when checks are green if I does not turn green before end of workday :)

PR #196, opened 2017-05-05
add support for specifying reservation to submit to via --job-reservation

### LINKED ISSUE #147: Support -W x=FLAGS:ADVRES:<reservation-name> flag
Support -W x=FLAGS:ADVRES:<reservation-name> flag. This is used to submit a hod job to a specific reservation so moab doesn't need to figure it out otherwise.

boegel:
fixes #147

### DIFF
--- a/docs/Command_line_interface.rst
+++ b/docs/Command_line_interface.rst
@@ -371,7 +371,7 @@ The number of cores per workernode to request; by default: ``-1``, i.e. full wor
 
 ``--job-reservation <reservation ID>``
 
-Reservation ID to submit job to (equivalent with using ``-W x=FLAGS:ADVRES:...`` with ``qsub``).
+Reservation ID to submit job into (equivalent with using ``-W x=FLAGS:ADVRES:<reservation ID>`` with ``qsub``).
 
 
 .. _cmdline_job_options_queue:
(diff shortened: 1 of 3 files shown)


boegel:
@stdweird please review

docs/Command_line_interface.rst
```diff
@@ -367,6 +367,13 @@ The number of (full) workernodes to request for the job being submitted (default
 The number of cores per workernode to request; by default: ``-1``, i.e. full workernodes (request all available cores).
 
 
+.. _cmdline_job_options_reservation:
+
+``--job-reservation <reservation ID>``
+
+Reservation ID to submit job to (equivalent with using ``-W x=FLAGS:ADVRES:...`` with ``qsub``).
```
stdweird: clarify a bit with `-W x=FLAGS:ADVRES:<resid>`? (or are they supposed to add the `FLAGS:ADVRES:` part?)
otoh, i'm not sure if users need to know this. don't we tell them to set some magic env varaible that is picked up by submitfilter?

stdweird:
LGTM. minor remark wrt to documnetation. feel free to correct and merge

docs/Command_line_interface.rst
boegel: No, we tell them to use `qsub -W x=FLAGS:ADVRES:<resid>`, so it makes sense to mention this in the documentation imho...

I'm not aware of the submitfilter support reservations, but that's a good idea, maybe we can make it smart enough to complain on non-existing reservations?

boegel: I suggest we follow-up on that in https://github.com/hpcugent/vsc-jobs/issues/97

stdweird:
@boegel please don't forget to make version bump PR

PR #202, opened 2019-01-24
add dist for Spark 2.4.0

boegel:


boegel:
Tested with a simple `Spark` example, works like a charm with the `Spark` installation obtained by the easyconfigs listed in https://github.com/easybuilders/easybuild-easyconfigs/pull/7485

